PR #12, opened 2025-05-20
Settings for ConfluenceAPI timeout and attachment limit per page

mjramer:
There is no configurable option for the Confluence API timeout or attachments limit, both are hardcoded to 75 seconds and 100, respectively. For large orgs with many nested wikis, fine-tuning the timeout and limit parameters is very important.

This PR aims to support environment variable flags CONFLUENCE_TIMEOUT and ATTACHMENT_LIMIT to allow users to tweak these settings.

Usage:

```
export CONFLUENCE_TIMEOUT=200
```
```
export ATTACHMENT_LIMIT=10
```


mjramer:
#11

mjramer:
@Spenhouet thoughts?

Spenhouet:
Hi @mjramer, thanks for your input. I see that making the timeout configurable is useful. I'm looking towards also making the retry settings and paging limits configurable as well.

But what is the use-case for limiting the attachments? As far as I can see, this will only lead to not all attachments being downloaded, which will result in incomplete conversions. I do see that it makes sense to implement paging here, as was done for the descendent pages. When paging is implemented, then configuring the limit does make sense.

Issue #10, opened 2025-05-20
Get Wikis by URL


mjramer:
We often are able to provide the wiki URL, not the page Id. We can use the confluence API get_page_by_title to retrieve the page ID, then trigger the Page.from_id workflow. We could change the arguments to support 'page-id' and 'page-url'. Happy to submit a PR as well.

Thoughts?

PR #13, opened 2025-05-20
Add support to get Wikis by URL

mjramer:
Often we don't know the page ID of wikis, and only know the URL. This PR adds support to get Wikis by URL.

Commands added:
```
confluence-markdown-exporter page-url <page-url>
```

Commands Modified:
```
confluence-markdown-exporter page-id <page-id>
```

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -17,12 +17,26 @@
       }
     },
     {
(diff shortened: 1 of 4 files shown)


mjramer:
#10

PR #23, opened 2025-05-22
Non-attachment image downloading

kitherill:


confluence_markdown_exporter/confluence.py
```diff
@@ -872,18 +887,122 @@ def convert_li(self, el: BeautifulSoup, text: str, parent_tags: list[str]) -> st
...
+                    # Use the first existing file with this hash
+                    file_path = existing_files[0]
+                else:
+                    # File not found, download it
+                    try:
+                        # Download the file
+                        url = confluence.url + src if src.startswith("/") else src
+                        
+                        # Use existing confluence session for other URLs
+                        response = confluence._session.get(url)
```
kitherill, line 965: Heads-up: if the session ends up having retrial enabled (there's a PR), this might be an overkill for the image downloading (in my case I had many 404/host name resolution errors). In this case it's better to use `requests.get()`, but it will lack the Atlassian auth.

PR #34, opened 2025-06-18
Optionally generate ADO style filenames

hputman-icrossing:
When Migrating articles from confluence to ADO I wanted the files names to match [ADO Wiki filenaming restrictions](https://learn.microsoft.com/en-us/azure/devops/organizations/settings/naming-restrictions?view=azure-devops#wiki-page-and-file-names). 

Created new Config setting:

```
PAGE_FILENAME_ADO=true
```
will generate md files with names based on page titles with dashes instead of spaces.  So page titled **Migrate & convert to Azure dev-ops** ends up in a file named `Migrate-&-convert-to-Azure-dev%2Dops.md`


Spenhouet:
Maybe it's possible to add more generic options to influence what filename sanitization is performed. Might even be multiple setting options, which combined give the "ADO-sanitization"

Spenhouet:
Sounds like a bit more finetuning but I like the approach!

hputman-icrossing:
Added escaping to the filename replace/encode patterns so `\`,`-`, `]`, or leading `^` in the pattern does not do unexpected things.

Spenhouet:
@hputman-icrossing Thanks for you work on this! 

I agree with your remark that the file ID should not contain any chars to sanitize, but then it should also not be an issue to do so? 

I found the current options of encoding and char replacement redundant and a bit confusing while we still kept ADO specific code.
I replaced the options with a single character mapping (encoding) which is applied. This should give the user full flexibility about what characters to replace with what. That also makes it easy to adopt for ADO specific logic.

Could you review the code? Does it cover your requirements? Or was something lost in my refactoring?

Spenhouet (approved):

PR #35, opened 2025-06-18
Optionally suppress breadcrumbs from generated page

hputman-icrossing:
When migrating a single page, you might be migrating into a different page hierarchy, so the breadcrumb trail from the original hierarchy doesn't make sense. This is a new option that can leave that breadcrumb trail out of the generated markdown.

```
PAGE_BREADCRUMBS=false
```

The value defaults to true to match current behavior.

### DIFF
--- a/confluence_markdown_exporter/confluence.py
+++ b/confluence_markdown_exporter/confluence.py
@@ -84,6 +84,11 @@ class ConverterSettings(BaseSettings):
             "  {attachment_extension} - Attachment file extension (including leading dot)"
         ),
     )
+    page_breadcrumbs: bool = Field(
+        default=True,
+        description="Whether to include breadcrumb links at the top of the page.",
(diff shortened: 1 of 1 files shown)


Spenhouet:
The Obsidian style already provides that option, but I see that it's much more useful to provide such settings individually, instead of trying to implement all sorts of export options as "ADO" or "Obsidian". This way, one can adjust to what ever target environment and the README can give some recommendations.

Spenhouet (approved):

Spenhouet:
@hputman-icrossing I removed the "style" specific configuration and changed to a feature flag like config. This is way more flexible. It's now possible to enable/disable breadcrumbs and document titles. I also added a section to the README on recommendations for systems like Obsidian and ADO. Please have a look. Any other config recommendation for ADO?

Thank you for your PR!

Issue #43, opened 2025-06-30
Allow multiple pages export


andren:
I'm testing this program in a very large Confluence organization (>4k items per space), where about <5 pages are of interest to me, thus the option to quickly export a small number of pages would be very helpful to me individually. I also think this would be helpful not just for regular users but also for development and testing. Here is what the command could look like:

`confluence-markdown-exporter pages 1 2 3`

PR #44, opened 2025-06-30
Add "pages" CLI command to export multiple pages

andren:
Addresses #43

# Before:
- Only possible to export entire spaces. It is not possible to export a small selection of pages by known ID or URL in a single command.

# After:
- Possible to export a smaller list of pages via the newly proposed "pages" argument:
`confluence-markdown-exporter pages 1 2 3`

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -22,7 +22,7 @@
       "request": "launch",
       "program": "${workspaceFolder}/confluence_markdown_exporter/main.py",
       "justMyCode": false,
-      "args": ["page", "<page-id> or <page-url>", "scratch"],
+      "args": ["page", "<page-id> or <page-url>", "--output-path", "scratch"],
       "console": "integratedTerminal",
       "cwd": "${workspaceFolder}",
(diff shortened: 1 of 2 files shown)


andren:
As a default to "page" command, certainly - I've pushed a new commit, let me know what you think.

There's a few questions one could raise about the UI (using plural for a single page? "Export pages 1 started at..." ; what if a large number of pages is provided?), but I've opted for implementation simplicity for now. We can definitely change that later if you prefer, just let me know.

-----
As for the "page_with_descendants" command - I'm not very experienced with Confluence, but I get the feeling that there might be potential for duplicate page exports in case two different pages share the same descendant (is this possible?). I'll think about this a bit more, maybe test it, and will get back to this thread later.

Spenhouet:
@andren Thanks for you contribution.
I agree that the commands should be plural.
I added the same feature to the "spaces" command.

There was an issue with the output path argument. If not provided, it would pick up the last page in the list.
I switched the output path to an option. This way this can not be confused.

Issue #47, opened 2025-07-04
Boolean environment variables are not evaluating "false" as a `False` value due to how Python handles truthiness


andren:
I was debugging a particular use-case irrelevant to this issue and my output was too verbose, so I wanted to set [DEBUG](https://github.com/Spenhouet/confluence-markdown-exporter/blob/5c1ea08754a4ce1eb5d7766dd687f4544440e180/.vscode/launch.json#L30) to `False` in the VSCode launch options.

It turns out that did not work, DEBUG still retained the value `True` because any non-empty string in Python is considered truthy. I was able to get `DEBUG=False` by setting `"DEBUG": ""` in the launch options.

Changing from `"DEBUG"="True"` to `"DEBUG"=""` does not feel intuitive.

This happened for env vars set through VSCode launch options, but it probably happens for any other way of setting env vars where the value is read as a string (terminal locally?, pipeline?, ...), and probably also for any other boolean env vars, not just DEBUG.

PR #48, opened 2025-07-04
Refactored DEBUG var to consider more intuitive boolean values

andren:
Addresses #47

This seems to be a common issue in Python, and I went down a little rabbit hole trying to find a good, common solution to this. See references at the bottom.

In the end, in between deprecated functions and some helper functions I've seen, I thought this approach was the simplest, but looking forward to feedback.

# Before
In [launch.json](https://github.com/Spenhouet/confluence-markdown-exporter/blob/5c1ea08754a4ce1eb5d7766dd687f4544440e180/.vscode/launch.json#L30):
```json
"env": {
  "PYTHONPATH": "${workspaceRoot}",
  "DEBUG": "false"
}
```

Results in:
`DEBUG=True`

# After
In [launch.json](https://github.com/Spenhouet/confluence-markdown-exporter/blob/5c1ea08754a4ce1eb5d7766dd687f4544440e180/.vscode/launch.json#L30):
```json
"env": {
  "PYTHONPATH": "${workspaceRoot}",
  "DEBUG": "false"
}
```

Results in:
`DEBUG=False`

---
# References
- https://danielms.site/zet/2023/pythons-distutil-strtobool-replacement/
- https://stackoverflow.com/questions/61361217/why-does-the-python-strtobool-method-return-an-int-instead-of-a-bool#:~:text=strtobool%20is%20a%20great%20function,loosely%20typed%20language%20like%20python.
- https://docs.python.org/2/distutils/apiref.html

### DIFF
--- a/confluence_markdown_exporter/api_clients.py
+++ b/confluence_markdown_exporter/api_clients.py
@@ -11,8 +11,9 @@ from confluence_markdown_exporter.utils.app_data_store import ApiDetails
 from confluence_markdown_exporter.utils.app_data_store import get_settings
 from confluence_markdown_exporter.utils.app_data_store import set_setting
 from confluence_markdown_exporter.utils.config_interactive import main_config_menu_loop
+from confluence_markdown_exporter.utils.type_converter import str_to_bool
 
-DEBUG: bool = os.getenv("DEBUG", "").lower() in {"true", "1"}
+DEBUG: bool = str_to_bool(os.getenv("DEBUG", "False"))
 
 
 def response_hook(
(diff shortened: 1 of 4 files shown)


andren:
One more thing, I also found `bool()` being used in [config_interactive.py](https://github.com/Spenhouet/confluence-markdown-exporter/blob/5c1ea08754a4ce1eb5d7766dd687f4544440e180/confluence_markdown_exporter/utils/config_interactive.py#L127).

I think there the value comes from a Python object, not a string, so I think issue #47 does not apply in that case, so I didn't make any changes there, but I did not explicitly test it.

Spenhouet:
@andren Thanks for the contribution!

Issue #51, opened 2025-07-08
Export process stops with an unhandled exception if the user account does not have permissions for a single page


andren:
I'm trying to export multiple pages, for example:
1. `confluence-markdown-exporter pages 1 2 3`
or
2. `confluence-markdown-exporter space mySpace`

Taking example 1., if my confluence account does not have permission for page 2, but does have permission for pages 1 and 3, the export process will halt at page 2, resulting in only page 1 having been exported. These are the last few lines of the exception (redacted):

```
╰────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
HTTPError: com.atlassian.confluence.api.service.exceptions.PermissionException: User _____:_____ does not have view permissions on entity page:       
_____ v.170
```

andren:
@Spenhouet, I've re-tested this scenario in 3.0.0 and this issue still occurs.


<img width="1612" height="310" alt="Image" src="https://github.com/user-attachments/assets/c822628f-6fe7-4b3f-87a9-265c531925cd" />

<img width="1691" height="250" alt="Image" src="https://github.com/user-attachments/assets/3cd6e713-6675-4983-933b-beafb8af3cbc" />



The confluence API is sending out an HTTPError when the user does not have permissions, so that exception needs to be caught.

We can catch the exception at [line 500 in confluence.py](https://github.com/Spenhouet/confluence-markdown-exporter/pull/52/commits/0d665835ac6fd33eb46bef7168f31ae7888248e7#diff-eb044bebcdc72a00cc7bdb853e00b9de3d678eaa1af6adf5901e662121849059L500). If you're concerned about exception specificity we can also add checks for `error_msg` containing "PermissionException" and/or `statusCode` == 403, or something along those lines.

Spenhouet:
Looks like a mistake on my part. I did not intent to remove the HTTPError. We can add it back.

PR #52, opened 2025-07-08
Fix unhandled exception that stops export process when user acount does not have permission for a single page

andren:
Addresses #51

# Before:
See #51

# After:
When running `confluence-markdown-exporter pages 1 2 3` and assuming no permissions for page 2 - pages 1 and 3 will be exported and the following logs (redacted) are visible:

```
Export pages 1, 2, 3 started at 2025-07-08 __:__:__
WARNING: Could not access page with ID ___: com.atlassian.confluence.api.service.exceptions.PermissionException: User ___:___ does not have view permissions on entity page: 2 v.170
Skipping export for inaccessible page with ID 2
Export pages 1, 2, 3 ended at 2025-07-08 __:__:__
Export pages 1, 2, 3 took relativedelta(seconds=+__, microseconds=+___)
```
---
Additionally, launch.json was failing due to the recent change of commands to plural, I've fixed that as an add-on commit to this PR. It's not really within the strict scope of the issue, but I did not want to spam too many PRs. Please let me know if you prefer to split that commit instead or feel free to commit it directly yourself.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -22,7 +22,7 @@
       "request": "launch",
       "program": "${workspaceFolder}/confluence_markdown_exporter/main.py",
       "justMyCode": false,
-      "args": ["page", "<page-id> or <page-url>", "--output-path", "scratch"],
+      "args": ["pages", "<page-id> or <page-url>", "--output-path", "scratch"],
       "console": "integratedTerminal",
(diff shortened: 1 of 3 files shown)


Spenhouet:
Thanks for noticing and fixing my mistake on the launch config.

PR #54, opened 2025-07-10
Add verify_ssl setting

twolfvb:
Given that the ApiClientFactory takes the retry_config from settings.retry_config.model_dump(), adding this verify_ssl bool field allows us to skip ssl checking, which is passed to the Confluence API, and it in turn gets passed to the requests library.

Then just modifying the json with the settings is enough to set this as False, so I can skip this step.

This is necessary to be able of using this project in my corporate intranet.

Let me know if this is OK with you.

I do feel like it being added on RetryConfig isn't ideal, maybe renaming RetryConfig to ConnectionConfig might be better. If that sounds better to you, let me know and I can do the necessary changes to it.


Spenhouet:
@twolfvb Thanks for the contribution. I adjusted the name to ConnectionConfig.

PR #55, opened 2025-07-11
Fix StopIteration caused by empty attachment.file_id on Confluence Server

finchy:
This PR fixes two crashes (StopIteration) when exporting pages whose attachments lack a file_id (common on older Server instances).
	•	Adds id fallback in get_attachment_by_id() / get_attachment_by_file_id()
	•	convert_attachment_link() now leaves the original Confluence URL if metadata is still missing.
Verified against Confluence 8.5.4: full-space export completes with --download-attachments

### DIFF
--- a/confluence_markdown_exporter/confluence.py
+++ b/confluence_markdown_exporter/confluence.py
@@ -825,21 +825,30 @@ class Page(Document):
             return f"[{page.title}]({page_path.replace(' ', '%20')})"
 
         def convert_attachment_link(self, el, text: str, parent_tags):
+            """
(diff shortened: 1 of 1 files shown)


Spenhouet:
I made some additional changes. 
There were two things I reverted from your change:
- I removed the backwards compatibility to the old export settings, as I do not see the need
- I reintroduced (also) reading the attachment file-id from `data-media-id`, as that it where this is stored on Confluence Cloud.

I think it should still cover your Confluence Server needs. Please test it.
For now I will merge the change and release it as part of v3.0.0

Issue #73, opened 2025-08-08
"createpage.action" links in older confluence pages can lead to a "maximum recursion depth exceeded" error.


andren:
This is a non-trivial issue, it involves some HTML and Confluence-specific topics and only occurs for what seems to be older Confluence pages, so please bear with me.

When running the command `cf-export pages 1`, if this Confluence page is old enough  and contains a "createpage.action" link, then it is quite possible that we will encounter an error like this (redacted):
```
│ ❱ 814 │   │   │   │   │   return self.convert_a(fallback, text, parent_tags)  # type: ignore -                                                                                                                                           │
│   815 │   │   │   │   return f"[[{text}]]"                                                                                                                                                                                               │
│   816 │   │   │   if "page" in str(el.get("data-linked-resource-type")):                                                                                                                                                                 │
│   817 │   │   │   │   page_id = str(el.get("data-linked-resource-id", ""))                                                                                                                                                               │
│                                                                                                                                                                                                                                          │
│ ╭───────────────────────────────────────────────────────────────────────────────── locals ─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮ │
│ │          el = <a class="createlink" href="https://wiki._____confluence_endpoint_____.com/pages/createpage.action?spaceKey=_____space_key_____&amp;title=_____link_title_____&amp;linkCreation=true&amp;fromPageId=_____page_id_____" │ │
│ │               rel="nofollow">_____link_title_____</a>                                                                                                                                                                                │ │
│ │    fallback = <a class="createlink" href="https://wiki._____confluence_endpoint_____.com/pages/createpage.action?spaceKey=_____space_key_____&amp;title=_____link_title_____&amp;linkCreation=true&amp;fromPageId=_____page_id_____" │ │
│ │               rel="nofollow">_____link_title_____</a>                                                                                                                                                                                │ │
│ │ parent_tags = {'div', 'p', 'tbody', 'tr', '[document]', '_inline', 'td', 'table'}                                                                                                                                                    │ │
│ │        self = <confluence_markdown_exporter.confluence.Page.Converter object at 0x00000_________>                                                                                                                                    │ │
│ │        text = '_____link_title_____'                                                                                                                                                                                                 │ │
│ ╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯ │
```
[...]

PR #74, opened 2025-08-08
Added fallback differentiation to avoid recursion depth issues

andren:
Addresses #73

# Before
image

see issue above for more details

# After
No recursion error for old Confluence pages anymore


emillg:
I have the same problem and this fix works well

Issue #53, opened 2025-07-09
Userspace tilde character '~' leads to a 400 Bad Request in commands related to spaces


andren:
In Confluence, user-specific spaces have this format: `~username`.

When running `confluence-markdown-exporter space ~myusername`, I got the following error (redacted):

```
│ ╭──────────────────────────────────────────────── locals ────────────────────────────────────────────────╮      │
│ │ http_error_msg = '400 Client Error: Bad Request for url: https://________.atlassian.net/wiki/res'+79   │      │
│ │         reason = 'Bad Request'                                                                         │      │
│ │           self = <Response [400]>                                                                      │      │
│ ╰────────────────────────────────────────────────────────────────────────────────────────────────────────╯      │
╰─────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
HTTPError: 400 Client Error: Bad Request for url:
https://________.atlassian.net/wiki/rest/api/space/C:%5CUsers%5Cmyusername?expand=homepage
```

The expected URL should have been something like:
`https://________.atlassian.net/wiki/rest/api/space/~myusername?expand=homepage`

After that, I put a breakpoint at [Space.from_key(space_key)](https://github.com/Spenhouet/confluence-markdown-exporter/blob/main/confluence_markdown_exporter/main.py#L76) and I modified the value of `space_key` directly:
- Before: `space_key = 'C:\\Users\\myusername'`
- After: `space_key = '~myusername'`

After that, the export was successful. This leads me to believe that somewhere along the code path, '~' is assumed to be the user directory of the current system, and thus it gets replaced with `C:\\Users\\` (in Windows), mangling the URL.

This was tested on Windows only. The issue occurs both when running confluence-markdown-exporter through launch.json or CLI, and for any of its recent versions.

andren:
This seems to be caused by Powershell itself. When running cf-export through CMD this does not happen.

PR #80, opened 2025-08-22
Handle powershell tilde expansion

andren:
Addresses #53

# Before
See issue above
image

# After
image

Made it case-insensitive and used regex to account for drive letters other than C.


Spenhouet:
@andren Is this PR ready from your side?

andren:
@Spenhouet yes, I tested right before my last comment. I didn't notice any additional flaws in the meantime.

The only breaking scenario I can imagine, is if someone has a Confluence page whose name starts with `C:\Users\` or `z:\users\` any other drive letter, case-insensitive. In this case, cf-exporter might wrongly replace that title with a tilde, but I haven't tested this yet.

PR #82, opened 2025-08-22
Feature ha append page metadata

hacoding1:
added features to append page metadata to bottom of exported markdown page

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -128,5 +128,9 @@ dmypy.json
 # Pyre type checker
 .pyre/
 
+# branch feature_ha_append_page_metadata
+tmp
+appTest.py
+
 # End of https://www.gitignore.io/api/code,python
 
--- a/confluence_markdown_exporter/confluence.py
+++ b/confluence_markdown_exporter/confluence.py
@@ -342,7 +342,7 @@ class Page(Document):
             'version' : response.get('version', {}).get('number','1'),
             'lastModified' : response.get('version', {}).get('when',''),
             'lastModifiedBy' : response.get('version', {}).get('by', {}).get('username',''),
-            'lastModifiedWithinLast365Days': response.get('version', {}).get('when','')[10] >= (datetime.now()-timedelta(days=365)).strftime('%Y-%m-%d') if response.get('version', {}).get('when','') else False,
+            'lastModifiedWithin365Days': response.get('version', {}).get('when','')[10] >= (datetime.now()-timedelta(days=365)).strftime('%Y-%m-%d') if response.get('version', {}).get('when','') else False,
             'pageURL' : f"""{response.get('_links', {}).get('base','')}/{response.get('_links', {}).get('webui','')}""",
             # 'pageURL' : f"""{response.get('_links', {}).get('base','')}/pages/viewpage.action?pageId={response.get('id','')}""",
         }
@@ -516,7 +516,7 @@ class Page(Document):
             ],
             attachments=Attachment.from_page_id(data.get("id", 0)),
             ancestors=[ancestor.get("id") for ancestor in data.get("ancestors", [])][1:],
-            page_metadata=data.get("page_metadata", {}),
+            page_metadata=data.get("page_metadata", ""),
         )
 
     @classmethod
@@ -548,7 +548,7 @@ class Page(Document):
                 labels=[],
                 attachments=[],
                 ancestors=[],
-                page_metadata='',
+                page_metadata="",
             )
 
     @classmethod
--- a/confluence_markdown_exporter/confluence.py
+++ b/confluence_markdown_exporter/confluence.py
@@ -16,6 +16,9 @@ from typing import Literal
 from typing import TypeAlias
 from typing import cast
 
+import json 
+from datetime import datetime, timedelta
+
 import jmespath
 import yaml
 from atlassian.errors import ApiError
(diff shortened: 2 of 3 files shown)


Spenhouet:
This PR requires a rebase and resolving merge conflicts.

Question: Why not integrate this page meta data via `set_page_properties`? We could have an option if the the page properties are printed as frontmatter or at the bottom of the page i.e. `page_properties_location=frontmatter|bottom`

ggramlich:
Hi @hacoding1 — quick heads-up: a sizeable portion of what you wanted to expose is now on `main` via #231 (merged today), and I'm preparing a follow-up PR to close the remaining gap.

**Already on `main`** (opt-in via `export.page_metadata_in_frontmatter: true`, default `false`):

| Your field | Now available as |
| --- | --- |
| `id` | `confluence_page_id` (string, JS-safe) |
| `version` | `confluence_version` |
| `lastModified` | `confluence_last_modified` |
| `lastModifiedBy` | `confluence_last_modified_by` |
| `pageURL` | `confluence_webui_url` / `confluence_tinyui_url` (via #229's `confluence_url_in_frontmatter`) |
| `title` | implicit (filename + optional H1 via `include_document_title`) |

Plus a bonus `confluence_space_key`.

**Follow-up PR I'm preparing** will add the still-missing fields to the same front-matter mechanism:

- `confluence_type` (page / blogpost)
- `confluence_created` (`history.createdDate`)
- `confluence_created_by` (`history.createdBy.displayName`)

The last two also need the `get_page_by_id` `expand=` to include `history`, which isn't fetched today.

Intentionally **not** porting `lastModifiedWithin365Days` — a derived flag whose value goes stale on the next read; downstream consumers can compute it from `confluence_last_modified` when they need it.

Re @Spenhouet's question about `page_properties_location=frontmatter|bottom`: the field collection in #231 is already factored such that a "render at bottom of page" mode could reuse the same dict. So once the missing fields land, your PR could realistically be reframed as the bottom-rendering option on top of the existing collector — no duplicated metadata logic.

ggramlich:
Quick follow-up to my [earlier note](https://github.com/Spenhouet/confluence-markdown-exporter/pull/82#issuecomment-4405850952): the follow-up PR is now open as **#242** — adds `confluence_type`, `confluence_created`, and `confluence_created_by` under the same `export.page_metadata_in_frontmatter` setting introduced by #231, no new toggle.

With #229, #231, and #242 (once merged) the only field from your original ten that wouldn't be available via front matter is `lastModifiedWithin365Days`, intentionally — derived flag, would go stale on the next export.

@hacoding1 — let me know if the three new keys cover what you needed; if there's a field on your original list I missed, easy to add to #242 before merge.

ggramlich:
@Spenhouet — separate from #242, picking up your `page_properties_location=frontmatter|bottom` thought from earlier. After landing #231 I had a closer look at what the current code already abstracts and what a "render the dict as a table" mode would actually need. Two viable shapes; happy to send the PR for whichever you prefer.

**Current state.** All four contributors to the YAML front matter (`labels` as tags, `_add_confluence_url_properties` from #229, `_add_page_metadata_properties` from #231 / #242, and `convert_page_properties` for the macro itself) write into the same `Page.Converter.page_properties` dict, which `front_matter` then dumps as YAML. The existing `page_properties_format` setting (`frontmatter`, `table`, `frontmatter_and_table`, `dataview-inline-field`, `meta-bind-view-fields`) only governs the position of the *original* Confluence macro table — not where the collected dict gets re-rendered. There is no "render the dict as a markdown table" path today, and no top/bottom slot in `Page.Converter.markdown` (which builds linearly: front matter → breadcrumbs → body). A markdown-table renderer (`tabulate(..., tablefmt="pipe")`) is already used inside `convert_page_properties` for `meta-bind-view-fields`, so the rendering primitive exists.

**Two design axes.** Independently of A vs. B below, I would split the decision into two orthogonal settings:

1. *What to render where:* `frontmatter`, `table`, `both`, or `none`.
2. *If a table is rendered, position relative to body:* `top` or `bottom` (default `bottom`, matching #82's original intent; `top` is useful for an "at-a-glance" header above breadcrumbs/body).

Keeping these orthogonal avoids a six-element literal and keeps each setting obvious.

**Option A — minimal, additive.** Widen `page_metadata_in_frontmatter` from `bool` to `Literal["none", "frontmatter", "table", "both"]` with a `model_validator` migration (`True → "frontmatter"`, `False → "none"`). Add `page_metadata_table_position: Literal["top", "bottom"] = "bottom"`. New helper `_render_page_metadata_table()` appends *only the page metadata fields* (the eight covered by #231 + #242) as a markdown table at the chosen position. URLs (#229) and labels stay front-matter-only.
Pros: touches only the metadata feature, no retroactive change to #229. Closes #82's actual ask — the original PR was specifically about page metadata at the bottom.
Cons: inconsistent — someone later wanting URLs in the same table would need a second pair of settings.

**Option B — your suggestion, taken global.** New setting `page_properties_location: Literal["frontmatter", "table", "both"] = "frontmatter"` that governs *all* dict-derived front-matter sources (labels-as-tags, URLs from #229, metadata from #231/#242) as a single switch. Plus `page_properties_table_position: Literal["top", "bottom"] = "bottom"`. The existing `*_in_frontmatter` settings remain "should this field be collected at all"; the new setting decides "where does the collected dict render". Implementation-wise: `_render_page_properties_table()` (singular, all keys) plus a top/bottom slot in `Page.Converter.markdown`.
Pros: one location switch, consistent across all three sources, matches your original phrasing.
Cons: retroactive surface-area change to two already-merged PRs; questions about granularity (labels are usually wanted as tags, not as a key/value table row — would `table` then exclude them? become a per-source allowlist?).

Which direction should the PR take? Happy to implement either; A is the smaller PR and unblocks #82's user-visible feature, B aligns with the longer-term shape you sketched. If you'd rather keep this open for now and just close #82 with a pointer to #231/#242, that's also a fine answer.

Spenhouet:
As this was not implemented by @ggramlich and merged to main (in parts not released yet), I will close this PR. @hacoding1 thanks for the initial work on this and @hputman-icrossing thanks for the support. @ggramlich Thanks for the final implementation.

Issue #79, opened 2025-08-21
Only images used in the currently exported page are downloaded as attachments


andren:
When exporting a page, only its visible images are downloaded as attachments.

It would be nice to have a setting that would define what kind of attachments would be downloaded.

## Use-case
the following page:
<img width="852" height="404" alt="Image" src="https://github.com/user-attachments/assets/fb196d2b-2a7b-4dbc-98b3-ad576f0d7c3e" />

has these attachments:
<img width="663" height="314" alt="Image" src="https://github.com/user-attachments/assets/1258ffbc-ba7b-4b83-9411-78889bc6dd7b" />

None of them are exported.

## Why?
[export_attachments()](https://github.com/Spenhouet/confluence-markdown-exporter/blob/40d7e28347162a81976409d7b46c137e64bc5b52/confluence_markdown_exporter/confluence.py#L434C1-L435C1) is what decides what gets exported or not.

In particular the line `if attachment.file_id in self.body:` and since the attachment id of those files is not mentioned in the body, they do not get exported. Thus only images visible in the Confluence page get exported.

## Proposal
Perhaps adding a setting that would allow to fetch:
1. Just visible images and .drawio and .drawio.png files (default and current functionality)
2. All available attachments

Spenhouet:
@andren Thanks for the suggestion and PR on this. Will be part of the next version.

PR #83, opened 2025-08-22
added setting to export all files regardless of reference in page

andren:
Addresses #79 

# Before:
See issue above
image

# After:
image
image
image

The Confluence space in the issue is the same as in this PR.


Spenhouet:
@andren Thanks for the suggestions and implementation. Looks like a good addition.

PR #113, opened 2025-11-12
Handle parsing of spaces where the homepage is None

Vyz333:
## Summary

This will make it so exports don't halt when there is a space with no homepage, as this happens sometimes.

## Test Plan

```
uv run pytest
uv run ruff check
```


andren:
I have the exact same issue - this contribution solves it for me. Thanks!

Spenhouet:
@Vyz333 Thanks for your contribution and @andren thanks for testing it. This will be part of the next release.

Note that this will better handle this case, but will not lead to an export of the space without homepage.

PR #128, opened 2026-01-17
Features to support better cooperation with Obsidian

mrtati:
## Summary
For my use case with Obsidian, I have added some features that the community might find useful as well.

These features are:
1. When a page or an attachment is downloaded, the last modified and last accessed time stamp of the downloaded file is set to the modification date from Confluence metadata (`version.when` value), this behavior is unconditional, should avoid unnecessary syncing and should help order files as they were edited on Confluence
2. Expanding the Table of Contents (TOC) macro was made optional by the new config flag `export.page_toc `, by default true (so no change to existing functionality while true)
3. New config flag `export.page_metadata_in_frontmatter` was introduced, by default disabled (no change to existing functionality while false), when enabled the front matter is filled with Confluence page metadata:
    - confluence_url: original url of the page, for easy clicking and visiting
    - confluence_version: version number
    - confluence_last_modified: last modification date of the downloaded page
    - confluence_last_editor: display name of the person who did the last change
4. New config flag `export.page_always_expand` was added, by default disabled (no change to existing functionality while false), when enabled the expand-container blocks are rendered as normal content without the `<details>\n<summary>` wrapping.

Additionally, there is a correction:
- Inline images were downloaded, but not shown in Obsidian. The reason seems to be that on our Confluence deployment, the image files and attachments do not have a `file_id`, only an `id` which is taken from a different html tag. Original implementation was kept, this functionality is additional, if the id is not found in `data-media-id`, it will be checked in `data-linked-resource-id` as well.

## Test Plan
Manual testing was done by changing the new config flags and observing the difference in output.


Spenhouet:
@mrtati There was a lot of change since you opened the PR. Most of your suggestions are implemented by now. Some are still missing like the page_always_expand option.

As this PR is pretty much outdated and would need redoing anyway, I will go ahead and close it. Feel free to open a new PR if any feature you'd like to see are still missing.

PR #153, opened 2026-03-11
fix parent_tags

khlkh:
## Summary

a) To handle our particular need where our Url always has a display
https://confluence.company.net/display/MySpace/MyPage
b) option to skip the embedded Jira items
c) add verbose and INFO to check progression as export took so long it appears to be stucked somewhere.
c) Fix some errors found during the run

## Test Plan

Mostly generated using AI. Not tested but it works for me.


Spenhouet:
Version 4.1.1 contains a fix for this issue. Please check it out.

PR #168, opened 2026-04-07
Fix _extract_base_url stripping context path for Server/DC instances

MathurinV:
## Summary

Server/Data Center instances deployed under a context path (e.g. `/confluence`) need that path preserved in the base URL for correct SDK endpoint resolution and auth config lookup.

The fix splits the URL path into segments and stops at the first known Confluence route segment (`wiki`, `display`, `spaces`, `rest`, `pages`, `plugins`, `dosearchsite.action`), keeping everything before it as the context path. Cloud URLs (`*.atlassian.net`) and API gateway URLs are unaffected.

Fixes #167

## Test Plan

- All 172 existing tests pass
- Verified manually against `https://host/confluence/spaces/KEY` → extracts `https://host/confluence`
- Verified Cloud URLs (`company.atlassian.net/wiki/spaces/KEY`) still extract `https://company.atlassian.net`
- Verified port preservation (`host:8090/confluence/display/PROJ`) → `https://host:8090/confluence`
- Verified nested context paths (`host/myapp/confluence/wiki/spaces/KEY`) → `https://host/myapp/confluence`

### DIFF
--- a/confluence_markdown_exporter/confluence.py
+++ b/confluence_markdown_exporter/confluence.py
@@ -74,22 +74,49 @@ _API_GATEWAY_HOST = "api.atlassian.com"
 def _extract_base_url(url: str) -> str:
     """Extract the base URL from a Confluence or Jira URL.
 
-    For standard instance URLs returns ``{scheme}://{hostname}``.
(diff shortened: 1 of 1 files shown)


MathurinV:
Looks good to me 😄

Issue #188, opened 2026-04-23
Page export fails for legacy Confluence Server/DC viewpage.action URLs with pageId query parameter


Zelayan:
### Description

When running cme page with a legacy Confluence Server/Data Center page URL like:
`cme page 'https://wiki.example.com/pages/viewpage.action?pageId=317425825'
`
the exporter fails before download starts.

The root cause is that this URL format is not parsed as a page-id based page URL. Instead, the path /pages/viewpage.action is incorrectly interpreted as:

space_key = "pages"
page_title = "viewpage.action"

This makes the tool call get_page_by_title(space="pages", title="viewpage.action"), which results in a 404 such as:
`No space with key : pages`

Depending on how the Atlassian client wraps the error, this may also surface as a misleading permission error.

Expected behavior:
legacy Confluence Server/DC URLs in the form .../pages/viewpage.action?pageId=... should resolve by pageId and export successfully.

Actual behavior:
the URL is misparsed as a space/title URL and export fails.

### Diagnostic info

```markdown

```

### Version

4.0.8

### Confluence Version

7.19.7

Spenhouet:
@Zelayan thanks for bringing this up and for providing a PR. This was merged with https://github.com/Spenhouet/confluence-markdown-exporter/pull/189 and will be part of the next release.

PR #189, opened 2026-04-23
fix:Add page ID resolution from URL query parameters in Page class

Zelayan:
## Summary
When running cme page with a legacy Confluence Server/Data Center page URL like:

cme page 'https://wiki.example.com/pages/viewpage.action?pageId=317425825'
the exporter fails before download starts.

The root cause is that this URL format is not parsed as a page-id based page URL. Instead, the path /pages/viewpage.action is incorrectly interpreted as:

space_key = "pages"
page_title = "viewpage.action"
This makes the tool call get_page_by_title(space="pages", title="viewpage.action"), which results in a 404 such as:

No space with key : pages
Depending on how the Atlassian client wraps the error, this may also surface as a misleading permission error.

Expected behavior:
legacy Confluence Server/DC URLs in the form .../pages/viewpage.action?pageId=... should resolve by pageId and export successfully.

Actual behavior:
the URL is misparsed as a space/title URL and export fails.


.github/workflows/ci.yml
Spenhouet, line 1: Please remove these changes of the uv setup version

confluence_markdown_exporter/confluence.py
Spenhouet, line 1: As this adds another supported page URL add this URL format in the README for supported URLs

Spenhouet (approved):

Spenhouet:
@Zelayan Thanks for your contribution.

Spenhouet:
Addressed https://github.com/Spenhouet/confluence-markdown-exporter/issues/188

Issue #150, opened 2026-02-27
Link to the original page in Confluence


efasel:
### Problem Description

Would be great if each page in the export would contain a link back to the original page in confluence.

### Proposed Solution

Add a link on each page which links back to the original page in confluence.

Spenhouet:
This was implemented as part of https://github.com/Spenhouet/confluence-markdown-exporter/pull/206
and will be part of the next version.

PR #206, opened 2026-04-30
Add optional Confluence page URL to exports

ahmedbektic:
Refs #150

## Summary

Adds attempt to build the Confluence URL from the Confluence API response's `_links.base` and `_links.webui` for all Page instantiations

Adds an optional export setting, `export.include_confluence_page_url`, that includes original Confluence URL in the exported md file like so:

    Confluence page: https://...

The setting defaults to `false` and has documentation.

Adds unit tests.

## Test Plan

- `uv run pytest`
- `uv run ruff check'
- Manual test: 
  - enabled `export.include_confluence_page_url`
  - manually exported a Confluence page, space, and org
  - verified exported Markdown files include the original Confluence page URL

### DIFF
--- a/README.md
+++ b/README.md
@@ -305,6 +305,13 @@ Whether to include breadcrumb links at the top of the page.
 - Default: `True`
 - ENV Var: `CME_EXPORT__PAGE_BREADCRUMBS`
 
+##### export.include_confluence_page_url
+
+Whether to include the original Confluence page URL at the top of the exported markdown file.
+
+- Default: `False`
+- ENV Var: `CME_EXPORT__INCLUDE_CONFLUENCE_PAGE_URL`
+
 ##### export.page_properties_as_front_matter
 
 Whether to convert Confluence Page Properties macro tables into YAML front matter. When enabled, key-value pairs in the macro are extracted and written as YAML front matter at the top of the exported file. When disabled, the macro is converted to a regular markdown table.
--- a/confluence_markdown_exporter/config.py
+++ b/confluence_markdown_exporter/config.py
@@ -32,6 +32,8 @@ _CONFIG_KEYS_EPILOG = (
     "| `export.attachment_href` | Link style for attachments: `relative` or `absolute` |\n\n"
     "| `export.include_document_title` | Prepend H1 title to each page |\n\n"
     "| `export.page_breadcrumbs` | Include breadcrumb links at top of page |\n\n"
+    "| `export.include_confluence_page_url` | Include original Confluence page URL "
+    "at top of page |\n\n"
     "| `export.enable_jira_enrichment` | Fetch Jira data for enriched links |\n\n"
     "| `export.attachment_export_all` | Export all attachments, not only referenced ones |\n\n"
     "| `export.filename_length` | Maximum filename length (default: 255) |\n\n"
--- a/confluence_markdown_exporter/confluence.py
+++ b/confluence_markdown_exporter/confluence.py
@@ -144,6 +144,25 @@ def _extract_base_url(url: str) -> str:
     return normalize_instance_url(base)
 
 
+def _get_web_url(data: JsonResponse) -> str:
+    links = data.get("_links", {})
+    if not isinstance(links, dict):
+        return ""
+
+    base = links.get("base")
+    webui = links.get("webui")
+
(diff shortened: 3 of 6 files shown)


Spenhouet:
@ahmedbektic Thanks for the PR. I'd prefer we add the page URL to the front matter via `self.set_page_properties` instead of adding it to the page content.

ahmedbektic:
> @ahmedbektic Thanks for the PR. I'd prefer we add the page URL to the front matter via `self.set_page_properties` instead of adding it to the page content.

I was not aware of this approach. However, I can see why you prefer it.

ahmedbektic:
@Spenhouet I could redo this (_if you'd like_), but do you have any other feedback?

Spenhouet:
The rest looks good to me. 👍

When you address that change request I can merge your change.

ggramlich:
I've been implementing something similar and will open a separate pull request.

My approach also covers both URL variants that the Confluence API provides via _links:

webui — the human-readable path including space key and page title (e.g. /wiki/spaces/MYSPACE/pages/12345/Page+Title). This is the canonical browser URL, but it changes when a page is renamed or moved.
tinyui — a short permalink based solely on the page ID (e.g. /wiki/x/AbCdEf). This remains stable across renames and moves, making it more suitable as a durable reference.
Both are combined with _links.base to form absolute URLs and exposed as source_url and source_url_tiny. They are written into the YAML front matter rather than the page body, which I believe aligns with the feedback given here.

Spenhouet:
@ggramlich happy to take your PR on this.

As a side note:

> a short permalink based solely on the page ID (e.g. /wiki/x/AbCdEf). This remains stable across renames and moves, making it more suitable as a durable reference.

These are not durable at all. They only live for a short period of time and then stop working. We forbid any usage of these URLs in our org.

ggramlich:
> @ggramlich happy to take your PR on this.
> 
> As a side note:
> 
> > a short permalink based solely on the page ID (e.g. /wiki/x/AbCdEf). This remains stable across renames and moves, making it more suitable as a durable reference.
> 
> These are not durable at all. They only live for a short period of time and then stop working. We forbid any usage of these URLs in our org.

Thanks for the feedback. Actually we don't use them that much when sharing page URLs, but are encouraged to use them. So no one was complaing about issues, as we barely use them.
I will have a 4-option literal config for choosing if and what to write in the frontmatter.

ggramlich:
@Spenhouet 
Question on coding style. Across the codebase, Literal[...] config fields are compared directly against raw strings at every call site (if settings.export.page_href == "wiki": ...). That's a valid convention, but it does make typos silent and renames manual.

Would you welcome a small refactor that introduces a shared symbol per Literal (e.g. a StrEnum) so the field type and its consumers share one source of truth? If you'd rather keep the current convention, just say so and I'll match it in any follow-up work.

Spenhouet:
Merged as part of https://github.com/Spenhouet/confluence-markdown-exporter/pull/206


@ahmedbektic thanks for the suggestion and work on this.

Spenhouet:
> f settings.export.page_href == "wiki": ..

@ggramlich the linter should capture a missmatch of a string against a Literal. Or is there a case where this does not work?

ggramlich:
> > f settings.export.page_href == "wiki": ..
> 
> @ggramlich the linter should capture a missmatch of a string against a Literal. Or is there a case where this does not work?

Fair point — and to be clear, I'm completely fine with the current style; I don't want to push a refactor on you. Just wanted to surface it for your call.

Concretely, with the lint stack from CONTRIBUTING.md (`ruff check` + `pytest`, no mypy/Pyright in CI), this passes silently:

```python
# Pydantic field:
attachments_export: Literal["referenced", "all", "disabled"]

# Consumer with a typo — "referencd" instead of "referenced":
if settings.export.attachments_export == "referencd":
    do_subset_export()
else:
    do_full_export()
```

The `if` branch is permanently dead, the `else` runs unconditionally — and `ruff check` is clean. With a shared symbol (`StrEnum` or constant), `AttachmentsExport.REFERENCD` would surface as an `AttributeError` the first time the line executes (so at least at test time, ideally in the editor). Whether that's worth a refactor in a small codebase is a judgement call — and absolutely yours to make.

PR #223, opened 2026-05-06
Add option to disable attachment download

ggramlich:
# PR: Add option to disable attachment download

## Title

```
Add option to disable attachment download
```

## Summary

Adds a new boolean setting `export.attachments_export` (default `True`, no
change in default behaviour). When set to `False`, `Page.export_attachments()`
returns immediately — no files are downloaded, no lockfile entries written, no
lockfile lookup performed.

**Why:** today the exporter always downloads attachments. For users who only
want the Markdown text (wiki-to-static-site pipelines, search/RAG ingestion,
quick text-only spot-exports of large spaces) this is wasted bandwidth, disk,
and — on big spaces — the dominant slice of total export time. The closest
existing setting, `export.attachment_export_all`, only changes *scope* (all
vs. referenced) and cannot turn the download off.

- Attachment **metadata** is still fetched in `Page.from_json` via
  `Attachment.from_page_id`. This is a cheap REST GET and keeps
  `Page.get_attachment_by_file_id` working, so image `src`s and file links in
  the Markdown body still resolve to the configured `attachment_path`. The
  files just won't exist locally.
- The `convert_attachments` macro still renders its Markdown table of links.
- `attachment_export_all` semantics, `Page.from_json`,
  `Page._attachments_for_export`, the body converters, and `Attachment.export`
  are untouched.

The flag is forward-only and does not retroactively delete previously
downloaded attachments.

The early-return is positioned **before** the lockfile lookup so disabled
pages produce zero stats noise and zero unnecessary I/O.

Both the README entry and the Pydantic field's `description=` text spell out
the trade-off (saves disk + bandwidth, but body links still point at
`attachment_path` even though the files don't exist locally). A new row is
added to the `_CONFIG_KEYS_EPILOG` table in `cme config` help output.

## Test Plan

Three new tests in `tests/unit/test_confluence.py` (class
`TestAttachmentsExportFlag`):

1. **Default (`True`) preserves behaviour:** `attachment.export()` is called
   and the returned dict contains the attachment entry.
2. **Disabled (`False`) skips download and lockfile lookup:**
   `attachment.export()` is **not** called, the returned dict is `{}`, and
   `LockfileManager.get_page_attachment_entries` is **not** called either —
   verifying the early-return is positioned before the lockfile read.
3. **Metadata still populated when disabled:** `Page.from_json` builds a Page
   with `len(page.attachments) == 1` even when `attachments_export=False`.
   Guards against future scope creep that would gate metadata loading on the
   same flag.

Automated:

```bash
uv run pytest tests/unit/ -q   # 252 passed
```

**Manually tested** against a live Confluence page that has at least one
attachment:

```bash
# Default — attachment files appear under output_path/.../attachments/
CME_EXPORT__OUTPUT_PATH=/tmp/cme-attach-on \
  uv run cme pages <url-with-attachment>
# Result: attachment files present, Markdown body links to them.

# Disabled — no attachment files are written
CME_EXPORT__OUTPUT_PATH=/tmp/cme-attach-off \
  CME_EXPORT__ATTACHMENTS_EXPORT=false \
  uv run cme pages <url-with-attachment>
# Result: no attachments/ directory created; Markdown body unchanged
# (image and file links still point at the configured attachment_path).
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -303,6 +303,13 @@ Export all attachments, not only those referenced by a page. Note: exporting lar
 - Default: `False`
 - ENV Var: `CME_EXPORT__ATTACHMENT_EXPORT_ALL`
 
+##### export.attachments_export
+
+Whether to download attachment files to disk. When disabled, no attachment files are written and no lockfile entries are created — useful when you only want the Markdown text and want to save disk space and bandwidth. Attachment metadata is still fetched from the Confluence API so image and file links in the page body continue to resolve in the rendered Markdown, but the referenced files will not exist locally.
+
+- Default: `True`
+- ENV Var: `CME_EXPORT__ATTACHMENTS_EXPORT`
+
 ##### export.image_captions
 
 Whether to export Confluence image captions in the exported Markdown. When enabled, the storage format of each page is fetched (via an additional API body expansion) and `ac:image` captions are extracted and rendered as an italic line directly below the image:
--- a/confluence_markdown_exporter/config.py
+++ b/confluence_markdown_exporter/config.py
@@ -40,6 +40,7 @@ _CONFIG_KEYS_EPILOG = (
     "| `export.convert_status_badges` | Convert Confluence status badges to `<mark>` elements |\n\n"
     "| `export.convert_text_highlights` | Convert background-color spans to `<mark>` elements |\n\n"
     "| `export.convert_font_colors` | Convert font-color spans to `<font>` elements |\n\n"
+    "| `export.attachments_export` | Download attachment files to disk (default: `true`) |\n\n"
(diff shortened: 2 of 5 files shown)


Spenhouet:
@ggramlich Thanks for the suggestion. I suggest we make this a Literal option with the options "referenced" (default), "all", "disabled" (or "none") and replace the old `attachment_export_all` with the new `attachments_export` config option.

In addition we should add a migration (see other migration related code) which sets the new config option to "referenced" when `attachment_export_all` was false and to "all" if `attachment_export_all` was true.

ggramlich:
@Spenhouet Thanks for the quick response. I understand and can prepare another pull request implementing the literal option and the migration for the config.

ggramlich:
Superseded by #228

Issue #39, opened 2025-06-19
Exported attachments only have suffix


cyanogenic:
When I export my space at Confluence Server 8.5.1 and I got this:

![Image](https://github.com/user-attachments/assets/fb191d0f-f1d5-4377-85db-3e0f2e706c7f)

Any ideas?

Spenhouet:
By default, attachments are stored at:
```sh
export ATTACHMENT_PATH="{space_name}/attachments/{attachment_file_id}{attachment_extension}"
```
Available variables:
  - `{space_key}`: The key of the Confluence space.
  - `{space_name}`: The name of the Confluence space.
  - `{homepage_id}`: The ID of the homepage of the Confluence space.
  - `{homepage_title}`: The title of the homepage of the Confluence space.
  - `{ancestor_ids}`: A slash-separated list of ancestor page IDs.
  - `{ancestor_titles}`: A slash-separated list of ancestor page titles.
  - `{attachment_id}`: The unique ID of the attachment.
  - `{attachment_title}`: The title of the attachment.
  - `{attachment_file_id}`: The file ID of the attachment.
  - `{attachment_extension}`: The file extension of the attachment, including the leading dot.

I don't have a Confluence Server to test. Maybe your Confluence Server does not provide an `attachment_file_id`.
You could try to set
```sh
export ATTACHMENT_PATH="{space_name}/attachments/{attachment_id}{attachment_extension}"
```
or
```sh
export ATTACHMENT_PATH="{space_name}/attachments/{attachment_title}{attachment_extension}"
```

cyanogenic:
`export ATTACHMENT_PATH="{space_name}/attachments/{attachment_title}{attachment_extension}"`
That works! Thanks a lot!

Spenhouet:
Great to hear!

ggramlich:
FYI: the root cause behind this was fixed in #224 — `{attachment_file_id}` now falls back to the content id on Confluence Server / Data Center, so the default `export.attachment_path` produces unique filenames without any override. The `{attachment_title}{attachment_extension}` variant suggested above remains a valid choice if you prefer human-readable filenames over numeric IDs.

PR #224, opened 2026-05-06
Fall back to content id when attachment fileId is empty

ggramlich:
# PR: Fall back to content id when attachment fileId is empty

## Title

```
Fall back to content id when attachment fileId is empty
```

## Summary

Fixes a long-standing issue where attachments collide on Confluence Data
Center / Server installations.

**The problem.** The default `export.attachment_path` template is
`{space_name}/attachments/{attachment_file_id}{attachment_extension}`. The
`{attachment_file_id}` placeholder is filled from `Attachment.file_id`, which
in turn comes from `extensions.fileId` in the Confluence REST response.
`extensions.fileId` is a Confluence-Cloud-specific GUID. **On Confluence Data
Center / Server, the API does not include this field**, so `file_id` becomes
an empty string for every attachment, every exported file collapses to a path
of the form `…/attachments/.<ext>`, and all attachments with the same
extension overwrite each other:

```
$ ls -la output/SOMESPACE/attachments/
-rw-rw-r-- 1 user user   266293 .docx
-rw-rw-r-- 1 user user    12138 .drawio
-rw-rw-r-- 1 user user 25203810 .m4a
-rw-rw-r-- 1 user user 35849567 .mp4
-rw-rw-r-- 1 user user 15850924 .pdf
…
```

DC/Server users currently have to manually override `attachment_path` to use
`{attachment_id}` to work around this. Cloud users are unaffected.

**The fix.** In `Attachment._template_vars`, fall back `attachment_file_id`
to the content id (`str(self.id)`, always populated on every Confluence
flavour) when `file_id` is empty:

```python
"attachment_file_id": self.file_id or str(self.id),
```

- Surgical change: only the template-variable derivation. The
  `Attachment.file_id` model field still represents what the API actually
  returned (empty string on DC), so anyone reading `file_id` directly sees no
  change.
- The default `attachment_path` template is **not** changed. With this fix,
  the existing default produces unique filenames on both Cloud and DC.
- No new setting, no migration required.

Both the README entry for `export.attachment_path` and the Pydantic field's
`description=` text now mention the DC/Server fallback.

**Out of scope:** migrating existing DC users' overridden `attachment_path`
configs (they keep working as-is); changing the default `attachment_path`
template (more controversial, separate change); sourcing a "real" file ID for
DC from somewhere else (e.g., from the download URL — the content id is
plenty unique).

## Test Plan

Three new tests in `tests/unit/test_confluence.py`:

1. **Cloud-style attachment (fileId present):** existing behaviour preserved
   — `attachment_file_id` is the GUID.
2. **DC-style attachment (fileId empty):** fallback kicks in —
   `attachment_file_id` is the content id.
3. **Two DC attachments with the same extension:** their `export_path` values
   are different (no collision).

Automated:

```bash
uv run pytest tests/unit/ -q   # 252 passed
```

**Manually tested** against a live Confluence Data Center instance with a
page that has multiple attachments of the same type. Before the fix, only one
file remained on disk; after the fix, every attachment is written to a
distinct path under `attachments/`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -296,6 +296,8 @@ Path template for attachments.
 - Default: `{space_name}/attachments/{attachment_file_id}{attachment_extension}`
 - ENV Var: `CME_EXPORT__ATTACHMENT_PATH`
 
+On Confluence Data Center / Server, where the API does not provide `fileId`, `{attachment_file_id}` falls back to the content id, so the default template still produces unique filenames.
+
 ##### export.attachment_export_all
 
 Export all attachments, not only those referenced by a page. Note: exporting large or many attachments increases export time.
--- a/confluence_markdown_exporter/confluence.py
+++ b/confluence_markdown_exporter/confluence.py
@@ -638,8 +638,11 @@ class Attachment(Document):
             **super()._template_vars,
             "attachment_id": str(self.id),
             "attachment_title": sanitize_filename(title_without_ext),
-            # file_id is a GUID and does not need sanitized.
-            "attachment_file_id": self.file_id,
+            # file_id is a GUID and does not need sanitization. On
+            # Confluence Data Center / Server the API does not populate
+            # fileId, so fall back to the content id which is always
+            # present and unique.
+            "attachment_file_id": self.file_id or str(self.id),
             "attachment_extension": self.extension,
         }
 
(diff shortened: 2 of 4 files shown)


Spenhouet (approved):

ggramlich:
Cross-reference: this fix also resolves the root cause behind #39. The workaround posted there — overriding `export.attachment_path` to use `{attachment_title}{attachment_extension}` or `{attachment_id}{attachment_extension}` — is the manual equivalent of what `{attachment_file_id}` now does automatically on Data Center / Server. The README's _Known Issues and Limitations_ entry that points to #39 is therefore obsolete and could be dropped in a small follow-up.

PR #228, opened 2026-05-06
Add option to control attachment download

ggramlich:
# PR: Add option to control attachment download

> Supersedes [#223](https://github.com/Spenhouet/confluence-markdown-exporter/pull/223). Same feature, redesigned per maintainer feedback ([comment](https://github.com/Spenhouet/confluence-markdown-exporter/pull/223#issuecomment-4386588123)) into a tri-state Literal that subsumes the existing `attachment_export_all` boolean.

## Title

```
Add option to control attachment download
```

## Summary

Replaces the boolean `export.attachment_export_all` with a single tri-state
`export.attachments_export: Literal["referenced", "all", "disabled"]`
(default `"referenced"` — preserves the previous default behaviour).

| Old `attachment_export_all` | New `attachments_export` | Behaviour |
|---|---|---|
| `False` (default) | `"referenced"` (default) | Export only attachments referenced in the page body |
| `True` | `"all"` | Export every attachment on the page |
| — | `"disabled"` | Skip downloads entirely — no files written, no lockfile entries, no lockfile lookup |

**Why.** Picks up the discussion from #223. The original PR introduced a
second boolean `attachments_export` to add a "disabled" mode. The maintainer
[suggested](https://github.com/Spenhouet/confluence-markdown-exporter/pull/223#issuecomment-4386588123)
collapsing both booleans into one Literal so users have a single, meaningfully
named knob for "which attachments to download" rather than two overlapping
flags. This PR implements that suggestion and adds the requested migration.

**What changes:**

- `ExportConfig.attachment_export_all: bool` is removed and replaced by
  `ExportConfig.attachments_export: Literal["referenced", "all", "disabled"]`.
- `Page._attachments_for_export()` now branches on
  `attachments_export == "all"` instead of the old boolean.
- `Page.export_attachments()` short-circuits with an empty result when
  `attachments_export == "disabled"`. The early-return is positioned **before**
  `LockfileManager.get_page_attachment_entries`, so disabled pages produce
  zero stats noise and zero unnecessary I/O.
- A new `@model_validator(mode="before")` migrates legacy configs:
  `attachment_export_all=False` → `"referenced"`,
  `attachment_export_all=True` → `"all"`. Mirrors the existing
  `_migrate_page_properties` migration in the same file.
  `description=` text are all updated to spell out the three modes and the
  trade-off in `"disabled"` mode (saves disk + bandwidth, but body image and
  file links still point at `attachment_path` even though the files don't
  exist locally).

- Attachment **metadata** is still fetched in `Page.from_json` via
  `Attachment.from_page_id` regardless of mode. This keeps
  `Page.get_attachment_by_file_id` working, so image `src`s and file links in
  the rendered Markdown body still resolve through the configured
  `attachment_path` even in `"disabled"` mode.
- The `convert_attachments` macro still renders its Markdown table of links.
- `Attachment.export` and the body converters are untouched.

The feature is forward-only: it does not retroactively delete previously
downloaded attachments.

**Migration behaviour (in-memory vs. on-disk).** The validator runs at config
load time and rewrites the legacy key to the new one **in memory**. The
on-disk `app_data.json` is *not* proactively rewritten — it stays with
`attachment_export_all` until the user next runs any command that saves the
config (`cme config set ...`, `cme config reset`, etc.), at which point
Pydantic's `model_dump_json()` only serialises the new field and the legacy
key disappears. This matches the behaviour of the existing
`_migrate_page_properties` migration; nothing breaks in the meantime.

**Note on env vars.** Pydantic Settings drops env vars that don't match a
declared field, so the migration does **not** apply to the legacy
`CME_EXPORT__ATTACHMENT_EXPORT_ALL` env var — users with that in CI need to
rename it to `CME_EXPORT__ATTACHMENTS_EXPORT` (with values
`referenced` / `all` / `disabled`).

## Test Plan

Updated tests in `tests/unit/test_confluence.py`:

- `TestAttachmentsForExport` (7 tests): each existing `attachment_export_all`
  case rewritten against the new Literal field
  (`"referenced"` for the body-reference subset, `"all"` for the
  return-everything case).
- New `TestAttachmentsExportFlag` (3 tests): default `"referenced"` still
  exports referenced attachments; `"disabled"` skips downloads **and** the
  lockfile lookup (verified by asserting `LockfileManager.get_page_attachment_entries`
  is not called); `Page.from_json` still populates `Page.attachments` when
  downloads are disabled (guards against future scope creep).

New tests in `tests/unit/utils/test_app_data_store_env.py`:

- `TestAttachmentsExportMigration` (3 tests): legacy
  `attachment_export_all=False` migrates to `"referenced"`; legacy
  `attachment_export_all=True` migrates to `"all"`; explicit new value wins
  over legacy when both are present (idempotent guard).

The existing `test_attachment_export_all_env_override` is replaced by
`test_attachments_export_env_override`, which verifies the new env var
`CME_EXPORT__ATTACHMENTS_EXPORT=all`.

Automated:

```bash
uv run pytest tests/unit/ -q   # 319 passed
```

**Manually tested** against a live Confluence page that has at least one
attachment:

```bash
# referenced (default) — only attachments referenced in body are downloaded
CME_EXPORT__OUTPUT_PATH=/tmp/cme-ref \
  uv run cme pages <url-with-attachment>

# all — every attachment on the page is downloaded
CME_EXPORT__OUTPUT_PATH=/tmp/cme-all \
  CME_EXPORT__ATTACHMENTS_EXPORT=all \
  uv run cme pages <url-with-attachment>

# disabled — no attachments/ directory created
CME_EXPORT__OUTPUT_PATH=/tmp/cme-off \
  CME_EXPORT__ATTACHMENTS_EXPORT=disabled \
  uv run cme pages <url-with-attachment>
find /tmp/cme-off -type d -name attachments  # empty
```

Migration smoke test (legacy config file):

```bash
mkdir -p /tmp/cme-legacy
cat > /tmp/cme-legacy/app_data.json <<'EOF'
{"export":{"attachment_export_all":true}}
EOF
CME_CONFIG_PATH=/tmp/cme-legacy/app_data.json \
  uv run python -c "from confluence_markdown_exporter.utils.app_data_store import get_settings; print(get_settings().export.attachments_export)"
# expect: all
```


Spenhouet (approved):

Spenhouet:
@ggramlich Thanks for this PR and for addressing my remark.

PR #229, opened 2026-05-06
Add Confluence page URL to YAML front matter

ggramlich:
# PR: Add Confluence page URL to YAML front matter

> Picks up the open feature work from [#206](https://github.com/Spenhouet/confluence-markdown-exporter/pull/206). Same idea — original Confluence page URL in exported markdown — but rebuilt around the maintainer's front matter request and extended to expose both URL variants the API provides.

## Title

```
Add Confluence page URL to YAML front matter
```

## Summary

Adds a single new setting,
`export.confluence_url_in_frontmatter: Literal["none", "webui", "tinyui", "both"]`
(default `"none"` — no behaviour change unless opted in), that writes the
original Confluence page URL into the YAML front matter of each exported
markdown file.

| Mode     | Front matter keys added                                              |
| -------- | -------------------------------------------------------------------- |
| `none`   | — (default; nothing changes)                                         |
| `webui`  | `confluence_webui_url` — human-readable URL (changes on rename/move) |
| `tinyui` | `confluence_tinyui_url` — stable short permalink based on page ID    |
| `both`   | both of the above                                                    |

**Why.** Picks up the discussion from [#206](https://github.com/Spenhouet/confluence-markdown-exporter/pull/206). The original PR appended the URL as a plain-text line to the body; the maintainer asked for the front matter via the existing `set_page_properties` mechanism instead. This PR implements that and also exposes the `tinyui` permalink — `webui` URLs encode the space key and current page title and break on rename or move, while `tinyui` paths (`/wiki/x/<base62>`) only encode the immutable page ID and survive both. Letting the user pick `webui` / `tinyui` / `both` keeps the choice in their hands without forcing one trade-off.

**What changes:**

- Three new helpers in `confluence.py` — `_join_confluence_link`,
  `_get_web_url`, `_get_tiny_url` — extract `_links.base` plus
  `_links.{webui|tinyui}` defensively (return `""` if either side is
  missing or the wrong type).
- `Page` gains two new fields, `web_url: str = ""` and `tiny_url: str = ""`,
  populated in `Page.from_json`. The existing fallback constructors in
  `Page.from_id` keep working unchanged because the defaults are empty
  strings.
- `Page.Converter` gains a new `_add_confluence_url_properties()` method
  that is called from the existing `front_matter` property right after
  `set_page_properties(tags=...)`. It writes the URLs through the same
  `page_properties` dict that the rest of the front matter machinery
  already uses.
- New Pydantic field
  `ExportConfig.confluence_url_in_frontmatter`. Modelled after the
  existing `page_properties_format` Literal (same docstring style, same
  default-disabled discipline).
  `description=` text are all updated to document the four modes.

**Collision safety.** If a Page Properties macro on the page already
populated `confluence_webui_url` or `confluence_tinyui_url` (for
example because the page author maintains the URL by hand inside a
Page Properties macro), the macro value wins —
`_add_confluence_url_properties` checks
`if key not in self.page_properties` before assigning. This is the only
reason it bypasses `set_page_properties()` and writes the dict
directly: that helper would unconditionally overwrite.

- The body of the page — no plain-text URL line is injected anywhere.
  PR #206's fallback approach is dropped.
- Default behaviour — `"none"` is the default, so existing exports are
  bit-identical.
- Internal raw-path fields some downstream forks expose (`webui_link`,
  `tinyui_link`) are not added; the public surface is just the resolved
  `web_url` / `tiny_url` strings on `Page`.
- No env-var migration: the feature is new; there is nothing legacy to
  migrate.

> Implementation written by Claude Sonnet 4.6 from a hand-written plan;

## Test Plan

(`TestConfluenceUrlInFrontmatter`, 12 tests):

- URL helpers: `_get_web_url` / `_get_tiny_url` join `_links.base` with
  the relative path; trailing-slash + leading-slash combinations are
  normalised; return `""` when `_links` is missing, when it isn't a
  dict, or when either side of the join is missing.
- Front matter: `webui` / `tinyui` / `both` modes each emit only the
  expected key(s); `none` emits neither; empty `Page.web_url` /
  `Page.tiny_url` short-circuit cleanly without writing empty values.
- Macro precedence: when `page_properties["confluence_webui_url"]` is
  already populated (simulating a Page Properties macro hit), the API
  URL is **not** injected and the existing value is preserved.

New test in `tests/unit/utils/test_app_data_store_env.py`:

- `test_confluence_url_in_frontmatter_env_override`: verifies
  `CME_EXPORT__CONFLUENCE_URL_IN_FRONTMATTER=both` reaches
  `settings.export.confluence_url_in_frontmatter`.

Automated:

```bash
uv run pytest tests/unit/ -q   # 337 passed
```

**Manually tested** against a live Confluence page:

```bash
# default — no URL in front matter
CME_EXPORT__OUTPUT_PATH=/tmp/cme-none \
  uv run cme pages <page-url>
grep -E "confluence_(webui|tinyui)_url:" /tmp/cme-none/**/*.md
# (no hits)

# both — both URLs land in front matter
CME_EXPORT__OUTPUT_PATH=/tmp/cme-both \
  CME_EXPORT__CONFLUENCE_URL_IN_FRONTMATTER=both \
  uv run cme pages <page-url>
grep -E "confluence_(webui|tinyui)_url:" /tmp/cme-both/**/*.md
# confluence_webui_url:  https://<host>/wiki/spaces/<KEY>/pages/<id>/<Title>
# confluence_tinyui_url: https://<host>/wiki/x/<short>
```

Macro-precedence smoke test: a page with a Page Properties macro whose
header sanitises to `confluence_webui_url` must export the macro value,
not the API URL.

### DIFF
--- a/README.md
+++ b/README.md
@@ -361,6 +361,22 @@ Controls how Confluence Page Properties Report macros (dynamic cross-page proper
 - Default: `frozen`
 - ENV Var: `CME_EXPORT__PAGE_PROPERTIES_REPORT_FORMAT`
 
+##### export.confluence_url_in_frontmatter
+
+Whether to include the original Confluence page URL in the YAML front matter of the exported file.
+
+| Value    | Description                                                                                                |
+| -------- | ---------------------------------------------------------------------------------------------------------- |
+| `none`   | Do not include any URL (default)                                                                           |
+| `webui`  | Include `confluence_webui_url` (human-readable URL; may change when the page is renamed or moved)          |
+| `tinyui` | Include `confluence_tinyui_url` (stable short permalink based on the page ID; survives renames and moves)  |
+| `both`   | Include both fields                                                                                        |
+
+If a Page Properties macro on the page already defines `confluence_webui_url` or `confluence_tinyui_url`, the value from the macro takes precedence over the URL extracted from the API.
+
+- Default: `none`
+- ENV Var: `CME_EXPORT__CONFLUENCE_URL_IN_FRONTMATTER`
+
 ##### export.filename_encoding
 
 Character mapping for filename encoding.
--- a/confluence_markdown_exporter/config.py
+++ b/confluence_markdown_exporter/config.py
@@ -33,6 +33,8 @@ _CONFIG_KEYS_EPILOG = (
     "| `export.include_document_title` | Prepend H1 title to each page |\n\n"
     "| `export.include_toc` | Export Table of Contents macro (`true`/`false`) |\n\n"
     "| `export.page_breadcrumbs` | Include breadcrumb links at top of page |\n\n"
+    "| `export.confluence_url_in_frontmatter` | Include Confluence page URL in YAML "
+    "front matter: `none`, `webui`, `tinyui`, `both` |\n\n"
     "| `export.enable_jira_enrichment` | Fetch Jira data for enriched links |\n\n"
     "| `export.attachments_export` | Which attachments to download:"
     " `referenced` (default), `all`, `disabled` |\n\n"
--- a/confluence_markdown_exporter/confluence.py
+++ b/confluence_markdown_exporter/confluence.py
@@ -164,6 +164,25 @@ def _extract_base_url(url: str) -> str:
     return normalize_instance_url(base)
 
 
+def _join_confluence_link(data: JsonResponse, key: str) -> str:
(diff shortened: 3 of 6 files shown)


Spenhouet:
@ggramlich Thanks for the PR.

PR #231, opened 2026-05-06
Feature/page metadata in frontmatter

ggramlich:
# PR: Add Confluence page metadata to YAML front matter

> Builds on the front matter mechanism merged in [#229](https://github.com/Spenhouet/confluence-markdown-exporter/pull/229). Same opt-in discipline (`confluence_*` keys, neutral default), but exposing five page metadata fields instead of URLs. Picks up the open requests from [#128](https://github.com/Spenhouet/confluence-markdown-exporter/issues/128) and [#82](https://github.com/Spenhouet/confluence-markdown-exporter/issues/82) (both closed/stale).

## Title

```
Add Confluence page metadata to YAML front matter
```

## Summary

Adds a single new setting,
`export.page_metadata_in_frontmatter: bool` (default `false` — no
behaviour change unless opted in), that writes five Confluence page
metadata fields into the YAML front matter of each exported markdown
file.

| Front matter key              | Source                              | Type   |
| ----------------------------- | ----------------------------------- | ------ |
| `confluence_page_id`          | `Page.id`                           | string |
| `confluence_space_key`        | `Page.space.key`                    | string |
| `confluence_last_modified`    | `Page.version.when` (ISO 8601)      | string |
| `confluence_last_modified_by` | `Page.version.by.display_name`      | string |
| `confluence_version`          | `Page.version.number`               | int    |

Fields with empty or zero values are omitted (e.g. an anonymous editor
with an empty `display_name` skips `confluence_last_modified_by` while
the other four fields still render).

**Why.** Picks up the same use case as [#128](https://github.com/Spenhouet/confluence-markdown-exporter/issues/128) and [#82](https://github.com/Spenhouet/confluence-markdown-exporter/issues/82) (both closed without a merged solution): downstream wiki tooling and static-site pipelines that need the exported markdown to carry stable identifiers (page ID, space key) and provenance (last-modified timestamp, author, version number). With the front matter machinery already in place from #229, the addition is small, additive, and matches the existing `confluence_*` naming convention.

**What changes:**

- New `Page.Converter._add_page_metadata_properties()` method, mirroring
  the existing `_add_confluence_url_properties()`. Called from
  `front_matter` immediately after the URL helper.
- New Pydantic field `ExportConfig.page_metadata_in_frontmatter`
  (`bool`, default `False`), placed directly after
  `confluence_url_in_frontmatter`.
  `description=` text are all updated.

**Type discipline.**

- `confluence_page_id` is intentionally serialized as a quoted string
  (`'629839369'`) rather than an integer. Confluence Cloud page IDs can
  exceed JavaScript's safe-integer range (`2^53 − 1`), so a JS-based
  static site generator (Hugo, Astro, …) consuming the front matter
  would silently truncate them. The Pydantic field on `Page` is `int`
  internally; only the front matter representation is coerced.
- `confluence_version` stays an integer in YAML (`7`, not `'7'`).
- `confluence_last_modified` is wrapped in single quotes by PyYAML's
  default emitter because the ISO-8601 string contains a timezone
  offset (`+02:00`); without quotes some YAML loaders would coerce it
  into a `datetime` object instead of preserving the exact string.

**Collision safety.** Same convention as
`_add_confluence_url_properties`: each key check uses
`if key not in self.page_properties`, so a Page Properties macro that
already populated `confluence_page_id` (or any of the other four keys)
on the page wins over the API-derived value.

- Default behaviour — `false` is the default, so existing exports are
  bit-identical.
- No env-var migration: the feature is new; there is nothing legacy to
  migrate.

> Implementation written by Claude Sonnet 4.6 from a hand-written plan;

## Test Plan

(`TestPageMetadataInFrontmatter`, 4 tests):

- `test_default_disabled_writes_no_metadata` — setting `False` (default):
  none of the five new keys appear in the front matter.
- `test_enabled_writes_all_five_keys` — setting `True`: all five keys
  present, with correct YAML types (`confluence_version` is `7`, not
  `'7'`; `confluence_page_id` is `'123'`, not `123`).
- `test_macro_precedence_for_page_id` — when `page_properties` already
  contains `confluence_page_id` (Page Properties macro hit), the
  API-derived value is **not** written.
- `test_empty_display_name_skipped` — empty `version.by.display_name`
  drops `confluence_last_modified_by`; the other four keys still render.

Existing `MockPage` test fixture extended with safe-default `space` and
`version` attributes so the new helper is callable from existing
`TestConfluenceUrlInFrontmatter` tests without breaking them.

New test in `tests/unit/utils/test_app_data_store_env.py`:

- `test_page_metadata_in_frontmatter_env_override`: verifies
  `CME_EXPORT__PAGE_METADATA_IN_FRONTMATTER=true` reaches
  `settings.export.page_metadata_in_frontmatter`.

Automated:

```bash
uv run pytest tests/unit/ -q   # 342 passed
```

**Manually tested** against a live Confluence page:

```bash
# default — no metadata in front matter
CME_EXPORT__OUTPUT_PATH=/tmp/cme-default \
  uv run cme pages <page-url>
grep -E "confluence_(page_id|space_key|last_modified|version):" /tmp/cme-default/**/*.md
# (no hits)

# enabled — all five fields land in front matter
CME_EXPORT__OUTPUT_PATH=/tmp/cme-meta \
  CME_EXPORT__PAGE_METADATA_IN_FRONTMATTER=true \
  uv run cme pages <page-url>
grep -E "confluence_(page_id|space_key|last_modified|last_modified_by|version):" /tmp/cme-meta/**/*.md
# confluence_page_id: '629839369'
# confluence_space_key: TEAM
# confluence_last_modified: '2026-04-17T13:25:54.491+02:00'
# confluence_last_modified_by: Alex Johnson
# confluence_version: 6
```

Macro-precedence smoke test: a page with a Page Properties macro whose
header sanitises to `confluence_page_id` must export the macro value,
not the API ID.


ggramlich:
**Schema mismatch with the inline-comments sidecar (#212) — happy to follow up either way**

Now that this PR adds `confluence_page_id` / `confluence_space_key` / etc. to the page front matter, there's a visible mismatch with the sidecar from #212, which uses unprefixed keys for partly the same metadata:

| | Page front matter (#229 + this PR) | Sidecar `*.comments.md` (#212) |
|---|---|---|
| Page identifier | `confluence_page_id: '123'` | `page_id: "123"` |
| Page title | (implicit — filename / `# heading`) | `page_title: "..."` |
| URL | `confluence_webui_url`, `confluence_tinyui_url` | `source: "{base}/wiki/spaces/{space}/pages/{id}"` |
| Quoting | PyYAML default | hand-written |

Two ways to resolve, each defensible:

1. **Align the sidecar to `confluence_*`** — e.g. `confluence_page_id`, `confluence_page_title`, plus a new `confluence_source_url` for the rename-stable URL the sidecar currently calls `source`. (`source` and `confluence_webui_url` aren't quite the same — `source` is title-stripped and rename-stable, `webui_url` carries the title slug — so it'd be a new key, not a rename.) Switching the sidecar to PyYAML emission would unify quoting as a side benefit.
2. **Keep the sidecar unprefixed by design** — on the grounds that the sidecar is a standalone document where there are no other front-matter sources to namespace against, so the prefix would just be noise. In that case the mismatch is intentional and worth a one-line README note so consumers don't trip over it.

Either way, this is a breaking change for one of the two file types, so it shouldn't ride along inside this PR. If you have a preference, I'm happy to follow up with a small dedicated PR in whichever direction you pick.

Spenhouet:
@ggramlich Would be great if you could align the sidecar to confluence_* as part of this PR. In this case I think the breaking change is acceptable as the comments export is a pretty new feature.

Spenhouet:
@ggramlich ping me if you think this is ready.

ggramlich:
@Spenhouet 
This should be ready for merge now

ggramlich:
@Spenhouet 
I can rebase again, if you want me to

Spenhouet:
@ggramlich thanks for the improvement

PR #232, opened 2026-05-07
Update Known Issues entry for missing attachment fileId

ggramlich:
## Summary

Follow-up to #224. That PR fixed the underlying problem from #39 by making `{attachment_file_id}` fall back to the content id on Confluence Server / Data Center, so the default `export.attachment_path` now produces unique filenames out of the box. The first entry under _Known Issues and Limitations_ in the README still describes the manual workaround as the solution, though, and is therefore out of date.

This PR rewrites that entry so it:

- mentions the automatic fallback (no override needed by default),
- recommends `{attachment_title}{attachment_extension}` as an alternative for users who prefer human-readable filenames over numeric IDs,
- keeps the link to #39 as historical context.

No code changes; documentation only. The default template is unchanged.

> Wording prepared by Claude Opus 4.7 from a hand-written plan; reviewed manually before opening this PR.

## Test Plan

- `git diff` shows only the one entry under _Known Issues and Limitations_ is changed.
- README renders cleanly on GitHub (Markdown list, inline code spans for template variables and example path).

### DIFF
--- a/README.md
+++ b/README.md
@@ -736,7 +736,7 @@ It generally was tested on:
 
 ## Known Issues and Limitations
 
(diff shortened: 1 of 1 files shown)


Spenhouet:
@ggramlich Thanks for the improvement

PR #241, opened 2026-05-08
Replace inline_comments bool with comments_export literal (closes #230)

### LINKED ISSUE #230: Extend inline comments sidecar to include page-level (footer) comments
## Context

PR #212 added `export.inline_comments`, which exports open inline comment
threads to a `<page>.comments.md` sidecar file. The Confluence REST API
exposes a second comment type via the same endpoint — page-level (footer)
comments — by passing `location="footer"` instead of `location="inline"`.

## Proposal

Add a separate opt-in setting, `export.page_comments: bool` (default
`False`), that fetches footer comments and includes them in the sidecar.
Like `inline_comments`, it would skip resolved comments and render replies
flat below their parent.

The two settings would be independent: you can export only inline comments,
only page comments, or both.

## Format question

When both settings are enabled, the sidecar could be structured in one of
two ways:

**Option A — single file, two sections:**
```markdown
---
page_id: "..."
page_title: "..."
source: "..."
---

## Inline comments

### <anchor text>
> quoted anchor
**Author** · 2026-04-01
Comment body…

## Page comments

### <first 60 chars of comment or "Comment <id>">
**Author** · 2026-04-01
Comment body…
```

**Option B — two separate sidecar files:**

- <page>.comments.md — inline (existing behaviour unchanged)
- <page>.page-comments.md — footer comments

Would you accept a PR for this? And if so, do you have a preference for
Option A or Option B?

### Alternatives Considered

_No response_

### Use Cases

_No response_

ggramlich:
## Summary

Replaces the existing boolean `export.inline_comments` with a single
literal setting `export.comments_export` that controls whether inline
and/or page-level (footer) comments are exported into the
`<page>.comments.md` sidecar.

| Value      | Behaviour                                                                                |
| ---------- | ---------------------------------------------------------------------------------------- |
| `none`     | No sidecar (default; behaviour-equivalent to current default).                           |
| `inline`   | Open inline comments only (annotated text as blockquote, then author / date / body).     |
| `footer`   | Open page-level (footer) comments only.                                                  |
| `all`      | Both, in a single sidecar with `## Inline comments` first, then `## Page comments`.      |

Resolved comments are skipped on both sides. Replies render flat below
their parent, identical to the existing inline-comments behaviour.

**Why.** Issue #230 asked whether footer comments should also land in
the sidecar. The owner picked Option A (single sidecar, two sections,
no second boolean) over a parallel `page_comments` flag. Folding the
existing boolean into a literal keeps the surface flat (one knob
instead of two) and matches the convention already used by
`export.attachments_export` and `export.confluence_url_in_frontmatter`
— literals over booleans whenever a third state becomes meaningful.

**What changes:**

- New Pydantic field `ExportConfig.comments_export:
  Literal["none", "inline", "footer", "all"]` (default `"none"`),
  replacing `inline_comments: bool`.
- New `model_validator(mode="before") _migrate_inline_comments`
  alongside the existing `_migrate_attachments_export`. Old
  `inline_comments=true/false` keys are silently mapped to
  `"inline"/"none"` and dropped from the dict, so a user upgrading
  picks up the new key without editing their config. Explicit
  `comments_export` always wins over a stale `inline_comments`.
- New `Page._fetch_page_comments()` paralleling `_fetch_inline_comments()`,
  with `location="footer"` and the resolved-comments filter applied
  identically.
- `Page.export_inline_comments()` renamed to
  `Page.export_comments_sidecar()` and split into
  `_render_inline_comments` / `_render_page_comments` helpers. Front
  matter (pinned by the regression test from #231) and the inline-body
  format are unchanged; the new combined output adds `## Inline comments`
  / `## Page comments` section headers above the per-comment blocks.
- `Page.export` call site and `Page.Converter.convert_inline_comment_marker`
  guard switched from the boolean to
  `comments_export in ("inline", "all")` (footer mode does not need
  marker recording — footer comments have no anchor text).
  `description=` are all updated. README adds a sidecar example showing
  both sections and a one-liner about the silent legacy-key migration.

**Migration discipline.**

- Migration is silent, matching the convention of the existing
  `_migrate_attachments_export`. No `DeprecationWarning`, no log line
  on startup. Happy to add a one-line `warnings.warn(...,
  DeprecationWarning, stacklevel=2)` in the validator if you'd prefer
  louder migration — call it out in review.
- The legacy key is consumed via `data.pop("inline_comments", None)`
  in the validator, so the old name does not survive on the model and
  cannot be re-read accidentally.

**What is intentionally not changed.**

- No `include_resolved` knob — resolved comments stay filtered out, in
  line with the convention since #212.
- No second sidecar file (`<page>.page-comments.md`) — Option A.
- No refactor of the front matter to `yaml.dump`. Out of scope for
  this PR; the existing string-templated front matter is byte-stable
  and pinned by the existing regression test.

## Test plan

(`TestPageCommentsSidecarBody`):

- `test_only_footer_writes_only_page_section` — `"footer"` writes the
  page section and nothing of the inline section.
- `test_inline_only_omits_page_section` — `"inline"` keeps the inline
  section header and excludes the page section even if footer
  comments are mocked.
- `test_all_writes_both_sections_inline_first` — `"all"` writes both
  sections; inline ordering enforced via `content.index(...)`.
- `test_none_writes_no_file` — `"none"` calls `save_file` zero times.
- `test_page_comment_title_falls_back_to_comment_id` — empty body
  picks the `Comment <id-prefix>` fallback (mirrors the inline-marker
  fallback).
- `test_page_comment_replies_render_under_parent` — two replies
  render in order under their parent comment.
- `test_fetch_page_comments_filters_resolved` — mocked HTTP client
  returns three comments (one resolved); `_fetch_page_comments` keeps
  the two open ones.

Updated tests:

- `TestInlineCommentsFrontMatter::test_front_matter_uses_confluence_prefix`
  switched from `Page.export_inline_comments(page)` to
  `Page.export_comments_sidecar(page)` and from
  `s.export.inline_comments = True` to
  `s.export.comments_export = "inline"`. The four `confluence_*` front
  matter keys remain pinned exactly as merged in #231.

Migration tests in `tests/unit/utils/test_app_data_store_env.py`
(`TestCommentsExportMigration`):

- `test_legacy_true_maps_to_inline` / `test_legacy_false_maps_to_none`
  — bool variants migrate to the right literal.
- `test_new_field_takes_precedence_over_old` — explicit new key wins
  over a stale legacy key.
- `test_legacy_key_does_not_appear_on_model` — legacy key is consumed,
  not re-exposed.
- `test_comments_export_env_override` —
  `CME_EXPORT__COMMENTS_EXPORT=all` flows through to
  `settings.export.comments_export`.

Automated:

```bash
uv run ruff check .            # All checks passed!
uv run pytest tests/ -q        # 398 passed
```

Manual smoke test (recommended before merge): pick a page with both
inline and footer comments and run the four modes against it,
verifying the produced `*.comments.md` sidecars match the modes:

```bash
for mode in none inline footer all; do
  rm -rf "/tmp/cme-comments-$mode"
  CME_EXPORT__OUTPUT_PATH=/tmp/cme-comments-$mode \
    CME_EXPORT__COMMENTS_EXPORT=$mode \
    uv run cme pages <page-url>
  echo "=== mode=$mode ==="
  ls /tmp/cme-comments-$mode/**/*.comments.md 2>/dev/null || echo "(no sidecar)"
done
```

---

This PR was drafted with assistance from an AI coding tool; all
changes have been reviewed locally before submission.


Spenhouet:
@ggramlich Thanks for that addition!

PR #242, opened 2026-05-08
Add type, created, created_by to page metadata front matter

ggramlich:
## Summary

Extends the existing `export.page_metadata_in_frontmatter: bool`
setting to also write three additional `confluence_*` keys:

| Front matter key         | Source                                  | Type   |
| ------------------------ | --------------------------------------- | ------ |
| `confluence_type`        | `Page.type` (`page` or `blogpost`)      | string |
| `confluence_created`     | `Page.history.created` (ISO 8601)       | string |
| `confluence_created_by`  | `Page.history.created_by.display_name`  | string |

Default behaviour is unchanged (`page_metadata_in_frontmatter: false`).
When enabled, the front matter now carries eight `confluence_*` keys
instead of five — full provenance for downstream wiki/SSG pipelines:
who created the page and when, alongside the existing last-modified
trio and the version number.

**Why.** [#82](https://github.com/Spenhouet/confluence-markdown-exporter/pull/82)
asked for ten page metadata fields rendered as a bottom-of-page table.
[#229](https://github.com/Spenhouet/confluence-markdown-exporter/pull/229)
covered the page URLs, [#231](https://github.com/Spenhouet/confluence-markdown-exporter/pull/231)
covered ID, space, version, and last-modified. The remaining gap is
the `history` block — *who created the page originally, and when* —
which downstream consumers need to distinguish original authorship
from the most recent editor (e.g. attribution in static-site
generators, audit reports, ownership rules in wiki tooling). Adding
three keys to the existing setting is additive, opt-in, and avoids a
second toggle.

`title` is intentionally not added: it is already the file name, and
optionally the H1 via `include_document_title`. `lastModifiedWithin365Days`
from #82 is also out of scope: it is a derived flag that would go stale
on the next export, and consumers can derive it from
`confluence_last_modified` themselves. Bottom-of-page rendering
(Spenhouet on #82, 2026-04-05: *"page_properties_location=frontmatter|bottom"*)
is its own discussion and a separate PR — this PR only extends the
existing front-matter mechanism.

**What changes:**

- New Pydantic model `History` in `confluence.py`, parallel to
  `Version`: `created: str`, `created_by: User`, with a `from_json`
  classmethod that reads `data["createdDate"]` and `data["createdBy"]`.
- `Page` gets `type: str = ""` and
  `history: History = Field(default_factory=...)`. The default-factory
  preserves the empty-page fallback path in `Page.from_id` for
  inaccessible pages without forcing every test that builds a `Page`
  via Pydantic to pass a `history`.
- `Page.from_json` populates both new fields from the API response.
- `Page.from_id` extends the `expand` string to
  `…,version,history,history.createdBy`, so the Confluence client
  actually returns the new sub-tree on Cloud and Server.
- `Page.Converter._add_page_metadata_properties()` writes the three
  new keys into the same `metadata` dict as before, in the order
  `page_id, space_key, type, created, created_by, last_modified,
  last_modified_by, version` (chronologically readable: identity → who
  made it → who touched it last → version).
- The existing filter (`if value in (None, "", 0): continue`) and
  collision check (`if key not in self.page_properties`) are unchanged,
  so a missing `type`, an anonymous original author, or a Page
  Properties macro that already populated any of the three new keys
  all behave consistently with the existing five.
  `description=` text all updated ("five" → "eight"; new rows in the
  field table; example block extended).

**No new setting.** Three additional fields under the existing
`page_metadata_in_frontmatter` toggle is additive and safe; a second
boolean would only add surface area for no real benefit.

**Type discipline.** `confluence_created` is wrapped in single quotes
by PyYAML's default emitter for the same reason
`confluence_last_modified` already is: the ISO-8601 string carries a
timezone offset (`+02:00`), and unquoted some YAML loaders would
coerce it into a `datetime` object instead of preserving the exact
string. README updated to mention both timestamps.

**Collision safety.** Same convention as before:
`if key not in self.page_properties` for each of the eight keys, so a
Page Properties macro entry for `confluence_type`, `confluence_created`,
or `confluence_created_by` wins over the API-derived value (parity
with the existing `confluence_page_id` precedence).

> Implementation drafted by Claude Opus 4.7 from a hand-written plan;

## Test Plan

Tests in `tests/unit/test_confluence.py`
(`TestPageMetadataInFrontmatter`):

- `test_default_disabled_writes_no_metadata` — extended: setting
  `False` keeps all eight keys (the original five plus
  `confluence_type`, `confluence_created`, `confluence_created_by`)
  out of the front matter.
- `test_enabled_writes_all_eight_keys` — replaces the earlier
  `_all_five_keys` test: setting `True` writes all eight keys with
  correct YAML types (`confluence_version` is `7`, not `'7'`;
  `confluence_page_id` is `'123'`, not `123`;
  `confluence_created` is single-quoted because of the `+02:00` offset).
- `test_blogpost_type_renders` — `Page.type = "blogpost"` lands as
  `confluence_type: blogpost`.
- `test_macro_precedence_for_page_id` — kept; alongside it a
  parametrised `test_macro_precedence_for_history_fields` covers
  `confluence_type`, `confluence_created`, `confluence_created_by`:
  when the Page Properties macro already populated any of the three,
  the macro value wins and the API value is **not** written.
- `test_empty_display_name_skipped` — kept; `test_empty_creator_skipped`
  added (anonymous original author drops `confluence_created_by` only)
  and `test_empty_type_skipped` added (empty `type` drops
  `confluence_type` only). The other seven keys still render in both
  cases.

`MockPage` fixture extended with `type=""` and a `history` MagicMock
carrying `created=""` and `created_by.display_name=""`, so existing
tests that build a `MockPage` continue to work without explicit setup.

The existing env-override test in
`tests/unit/utils/test_app_data_store_env.py` already covers
`page_metadata_in_frontmatter`; no new env test is needed since the
setting itself is unchanged.

Automated:

```bash
uv run pytest tests/           # 404 passed
```

**Manually tested** against a live Confluence page (Server instance,
real production page, anonymised display names and host below):

```bash
CME_EXPORT__OUTPUT_PATH=/tmp/cme-meta-history \
  CME_EXPORT__CONFLUENCE_URL_IN_FRONTMATTER=both \
  CME_EXPORT__PAGE_METADATA_IN_FRONTMATTER=true \
  uv run cme pages <page-url>

head -12 /tmp/cme-meta-history/**/*.md
# ---
# confluence_created: '2023-12-01T10:37:25.714+01:00'
# confluence_created_by: Sam Creator
# confluence_last_modified: '2025-09-10T13:34:37.412+02:00'
# confluence_last_modified_by: Alex Johnson
# confluence_page_id: '629839369'
# confluence_space_key: TEAM
# confluence_tinyui_url: https://example.atlassian.net/x/QwExHQ
# confluence_type: page
# confluence_version: 122
# confluence_webui_url: https://example.atlassian.net/wiki/spaces/TEAM/pages/629839369/REST-APIs
# ---
```

(YAML key order is alphabetical because PyYAML's default dump sorts
keys; the implementation builds the metadata dict in chronological
order — `page_id, space_key, type, created, created_by, last_modified,
last_modified_by, version` — and the existing front-matter emitter
sorts on output. Both orderings are stable; downstream consumers index
by key name, not position.)

Macro-precedence smoke test on a page whose Page Properties macro
header sanitises to `confluence_type`: the macro value (`policy`,
say) takes precedence over the API value (`page`), matching the
behaviour pinned by `test_macro_precedence_for_page_id` from #231.

## Out of scope

- **Bottom-of-page rendering** (Spenhouet's `page_properties_location`
  question on #82). Belongs in its own PR — this one is purely an
  additive front-matter extension.
- **`lastModifiedWithin365Days`** from #82. Derived flag, would go
  stale on the next export; consumers can compute it from
  `confluence_last_modified`.
- **A second toggle** for the history fields. Three additional keys
  under the existing setting is additive and safe; doubling the
  surface area would only add complexity for no real benefit.


Spenhouet:
@ggramlich thanks for this addition

PR #245, opened 2026-05-13
feat: add PlantUML support for Confluence Server/Data Center

kasadna:
## Summary

On Confluence Server and Data Center, `body.editor2` is often empty and PlantUML diagrams are stored as raw `@startuml`/`@enduml` source inside `body.storage` CDATA sections. The view HTML renders them as `<span data-macro-name="plantuml">` elements without a `data-macro-id`.

The existing `convert_plantuml` only supports the Cloud format (editor2 + macro-id matching), so Server/DC exports produce flattened SVG text or HTML comments instead of fenced PlantUML code blocks.

## Changes

- **Always include `body.storage`** in the REST API expand parameter (previously only included when `image_captions` was enabled)
- **Add `convert_span` dispatch** for `plantuml` macros — Server renders PlantUML inside `<span>` elements rather than `<div>`
- **Add body.storage fallback** in `convert_plantuml` using positional matching when editor2/macro-id is unavailable
- **Extract helper methods** `_extract_uml_from_editor2` and `_extract_uml_from_storage` for cleaner separation
- **Add `_storage_plantuml_macros`** cached property for parsing body.storage macros once per page

## How it works

1. **Cloud path (unchanged):** If `data-macro-id` is present, look up the matching macro in editor2 XML and parse the JSON `umlDefinition`.
2. **Server/DC path (new):** If editor2 lookup fails, fall back to positional matching — the Nth PlantUML element in `body.view` maps to the Nth `ac:structured-macro` in `body.storage`.

## Testing

- Updated existing PlantUML tests to cover both Cloud and Server formats
- Added tests for: Server storage extraction, multiple diagrams positional matching, editor2-to-storage fallback, invalid JSON fallback
- All 407 tests pass with `uv run pytest`
- All lint checks pass with `uv run ruff check`

## Real-world validation

Tested against a Confluence Server instance with 246 pages. 15 pages contained PlantUML diagrams that were previously exported as flattened SVG text — all now export correctly as fenced ```plantuml``` code blocks.

### DIFF
--- a/confluence_markdown_exporter/confluence.py
+++ b/confluence_markdown_exporter/confluence.py
@@ -1338,11 +1338,9 @@ class Page(Document):
             )
         logger.debug("Fetching page id=%s from %s", page_id, base_url)
         expand = (
-            "body.view,body.export_view,body.editor2,metadata.labels,"
+            "body.view,body.export_view,body.editor2,body.storage,metadata.labels,"
             "metadata.properties,ancestors,version,history,history.createdBy"
         )
-        if settings.export.image_captions:
-            expand += ",body.storage"
         try:
             return cls.from_json(
                 _require_dict(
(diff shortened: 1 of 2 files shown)


Spenhouet (approved):

Spenhouet:
@kasadna Thanks for the improvement

PR #246, opened 2026-05-13
feat: auto-detect PlantUML in code blocks

kasadna:
## Summary

Some Confluence pages embed PlantUML source inside Code Macros (with brush set to `java`, `text`, etc.) rather than using the dedicated PlantUML macro. These are exported as fenced code blocks with the wrong language tag, making them invisible to PlantUML renderers.

## Changes

A 2-line addition in `convert_pre`: if the code block text contains `@startuml`, override the fence language to `plantuml` regardless of the brush parameter.

## Testing

- Added `test_plantuml_code_block_detection.py` with 4 test cases:
  - Code block with `@startuml` + brush → uses `plantuml` fence
  - Regular code block without `@startuml` → keeps original language
  - Empty code block → returns empty
  - Code block without brush param but containing `@startuml` → gets `plantuml` fence
- All 408 tests pass with `uv run pytest`
- All lint checks pass with `uv run ruff check`

## Real-world validation

Tested against a Confluence Server instance with 246 pages. Several pages had PlantUML source embedded in Code Macros that are now correctly fenced as ```plantuml```.

### DIFF
--- a/confluence_markdown_exporter/confluence.py
+++ b/confluence_markdown_exporter/confluence.py
@@ -1938,6 +1938,9 @@ class Page(Document):
                 if match:
                     code_language = match.group(1)
 
+            if "@startuml" in text:
+                code_language = "plantuml"
+
             return f"\n\n```{code_language}\n{text}\n```\n\n"
 
         def convert_sub(self, el: BeautifulSoup, text: str, parent_tags: list[str]) -> str:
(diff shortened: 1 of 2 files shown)


Spenhouet (approved):

Spenhouet:
@kasadna Thanks for the addition

PR #257, opened 2026-05-21
feat: add SSO cookie authentication

UmbraCi:
Many internally deployed Confluence instances rely on SSO/session login, where API token authentication is not always available. This PR adds SSO cookie authentication support for the Confluence API client, extracted and cleaned up from #233 as requested by the maintainer.

PR #233, opened 2026-05-07
feat: add SSO cookie auth and nested-table Markdown sidecar export

UmbraCi:
## Background
Many internally deployed Confluence instances rely on SSO/session login, where API token authentication is not always available in practice.

At the same time, nested tables can lose structure fidelity during Markdown export.

## Goal
1. Support authenticated export in internal Confluence environments.
2. Preserve nested-table information in exported Markdown.

## Changes
- Added cookie-based authentication support for Confluence API client creation.
- Added dedicated handling for nested tables in Markdown export to keep table structure information.

## Expected Outcome
- The exporter can run in SSO-based internal deployments via session cookies.
- Exported Markdown retains nested-table information more reliably.

PR #258, opened 2026-05-21
feat: export high-fidelity nested table Markdown sidecar

UmbraCi:
### Description

This Pull Request addresses the visual fidelity issue when exporting complex Confluence pages containing nested tables. 

#### The Problem
Standard Markdown does not natively support nested tables. When `confluence-markdown-exporter` encounters a nested table structure, standard Markdown conversion flattens it, resulting in broken layouts, misaligned columns, or lost structural contexts.

#### The Solution
To preserve the exact visual fidelity of nested tables without breaking the standard Markdown output, we introduce a **high-fidelity sidecar export option**:

1. **`export.visual_fidelity_markdown` (boolean, default: `False`)**: When enabled, the exporter writes an additional visual-fidelity Markdown sidecar file next to each exported page.
2. **`export.visual_fidelity_suffix` (string, default: `\".visual\"`)**: Suffix inserted before the `.md` extension for the visual-fidelity sidecar file (e.g., `page.visual.md`).
3. **HTML Preservation**: In the sidecar file, nested tables are preserved exactly in their original HTML format, keeping the layout intact while rendering correctly in Markdown viewer applications (e.g., Obsidian).

This PR is part of a split of the original PR #233 into two separate, cleanly scoped PRs (with SSO Cookie Auth separated into PR #257). All configurations are fully covered by unit and integration tests, and environment variable overrides are supported.

### Verification
All unit tests pass successfully.


UmbraCi:
### 📊 Nested Table Export Comparison

Here is a real-world export comparison of a complex nested table extracted from Private deployment of Confluence using our local merged development build containing both SSO cookie authentication and high-fidelity nested table preservation logic.

---

#### 1️⃣ Standard Markdown Export (Flattened)
*In the standard export, the nested table structure is flattened. The inner table's cells are converted to Markdown lines with `<br/>` tags inside the outer cell, and the outer table columns are padded with empty cells. While readable, it breaks the visual fidelity of the original grid.*

```markdown
|                                                                                                                                     |                                                                                                                                              |                                                                                                    |                                                                      |     |     |     |     |     |   1 | 2          | 3    | 4       |
|:------------------------------------------------------------------------------------------------------------------------------------|:---------------------------------------------------------------------------------------------------------------------------------------------|:---------------------------------------------------------------------------------------------------|:---------------------------------------------------------------------|----:|----:|----:|----:|----:|----:|:-----------|:-----|:--------|
| \|   1.1 \|   1.2 \|   1.3 \|<br/>\|------:\|------:\|------:\|<br/>\|   1.4 \|   1.5 \|   1.6 \|<br/>\|   1.7 \|   1.8 \|   1.9 \| | 1.1                                                                                                                                          | 1.2                                                                                                | 1.3                                                                  | 1.4 | 1.5 | 1.6 | 1.7 | 1.8 | 1.9 | Background | Goal | Changes |
| 1.1                                                                                                                                 | 1.2                                                                                                                                          | 1.3                                                                                                |                                                                      |     |     |     |     |     |     |            |      |         |
| 1.4                                                                                                                                 | 1.5                                                                                                                                          | 1.6                                                                                                |                                                                      |     |     |     |     |     |     |            |      |         |
| 1.7                                                                                                                                 | 1.8                                                                                                                                          | 1.9                                                                                                |                                                                      |     |     |     |     |     |     |            |      |         |
| And please correct the branch name:                                                                                                 | Many internally deployed Confluence instances rely on SSO/session login, where API token authentication is not always available in practice. | Added cookie-based authentication support for Confluence API client creation.                      | 1. Support authenticated export in internal Confluence environments. |     |     |     |     |     |     |            |      |         |
| <font style="color: #ff0000;">That's an em dash (—) in your branch name, not a hyphen.</font>                                       | At the same time, nested tables can lose structure fidelity during Markdown export.                                                          | Added dedicated handling for nested tables in Markdown export to keep table structure information. | 1. Preserve nested-table information in exported Markdown.           |     |     |     |     |     |     |            |      |         |
```

---

#### 2️⃣ High-Fidelity Markdown Export (`.visual.md` Sidecar)
*In the high-fidelity sidecar export (`.visual.md`), the nested table's HTML block is completely preserved. The outer table is generated as standard Markdown, but the inner nested table's exact cells, rows, and columns are rendered using pure HTML `<table>` blocks inside the Markdown cells. Markdown viewer applications (e.g. GitHub, Obsidian, VS Code) will render the nested tables with 100% visual fidelity.*

```markdown
| 1                                                                                                                                                                                                                                                                                                                                                                                          | 2                                                                                                                                            | 3                                                                                                  | 4                                                                    |
|:-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|:---------------------------------------------------------------------------------------------------------------------------------------------|:---------------------------------------------------------------------------------------------------|:---------------------------------------------------------------------|
| <table class="confluenceTable"><tbody><tr><th class="confluenceTh">1.1</th><th class="confluenceTh">1.2</th><th class="confluenceTh">1.3</th></tr><tr><td class="confluenceTd">1.4</td><td class="confluenceTd">1.5</td><td class="confluenceTd">1.6</td></tr><tr><td class="confluenceTd">1.7</td><td class="confluenceTd">1.8</td><td class="confluenceTd">1.9</td></tr></tbody></table> | Background                                                                                                                                   | Goal                                                                                               | Changes                                                              |
| And please correct the branch name:                                                                                                                                                                                                                                                                                                                                                        | Many internally deployed Confluence instances rely on SSO/session login, where API token authentication is not always available in practice. | Added cookie-based authentication support for Confluence API client creation.                      | 1. Support authenticated export in internal Confluence environments. |
| <font style="color: #ff0000;">That's an em dash (—) in your branch name, not a hyphen.</font>                                                                                                                                                                                                                                                                                              | At the same time, nested tables can lose structure fidelity during Markdown export.                                                          | Added dedicated handling for nested tables in Markdown export to keep table structure information. | 1. Preserve nested-table information in exported Markdown.           |
```

---

#### 📌 Visual Preview inside GitHub / GFM Rendering

##### 1. Standard (Flattened - outer columns are offset and inner structure collapses):
|                                                                                                                                     |                                                                                                                                              |                                                                                                    |                                                                      |     |     |     |     |     |   1 | 2          | 3    | 4       |
|:------------------------------------------------------------------------------------------------------------------------------------|:---------------------------------------------------------------------------------------------------------------------------------------------|:---------------------------------------------------------------------------------------------------|:---------------------------------------------------------------------|----:|----:|----:|----:|----:|----:|:-----------|:-----|:--------|
| \|   1.1 \|   1.2 \|   1.3 \|<br/>\|------:\|------:\|------:\|<br/>\|   1.4 \|   1.5 \|   1.6 \|<br/>\|   1.7 \|   1.8 \|   1.9 \| | 1.1                                                                                                                                          | 1.2                                                                                                | 1.3                                                                  | 1.4 | 1.5 | 1.6 | 1.7 | 1.8 | 1.9 | Background | Goal | Changes |
| 1.1                                                                                                                                 | 1.2                                                                                                                                          | 1.3                                                                                                |                                                                      |     |     |     |     |     |     |            |      |         |
| 1.4                                                                                                                                 | 1.5                                                                                                                                          | 1.6                                                                                                |                                                                      |     |     |     |     |     |     |            |      |         |
| 1.7                                                                                                                                 | 1.8                                                                                                                                          | 1.9                                                                                                |                                                                      |     |     |     |     |     |     |            |      |         |
| And please correct the branch name:                                                                                                 | Many internally deployed Confluence instances rely on SSO/session login, where API token authentication is not always available in practice. | Added cookie-based authentication support for Confluence API client creation.                      | 1. Support authenticated export in internal Confluence environments. |     |     |     |     |     |     |            |      |         |
| <font style="color: #ff0000;">That's an em dash (—) in your branch name, not a hyphen.</font>                                       | At the same time, nested tables can lose structure fidelity during Markdown export.                                                          | Added dedicated handling for nested tables in Markdown export to keep table structure information. | 1. Preserve nested-table information in exported Markdown.           |     |     |     |     |     |     |            |      |         |

##### 2. High-Fidelity (Visual Sidecar - preserves nested structure seamlessly):
| 1 | 2 | 3 | 4 |
|:---|:---|:---|:---|
| <table class="confluenceTable"><tbody><tr><th class="confluenceTh">1.1</th><th class="confluenceTh">1.2</th><th class="confluenceTh">1.3</th></tr><tr><td class="confluenceTd">1.4</td><td class="confluenceTd">1.5</td><td class="confluenceTd">1.6</td></tr><tr><td class="confluenceTd">1.7</td><td class="confluenceTd">1.8</td><td class="confluenceTd">1.9</td></tr></tbody></table> | Background | Goal | Changes |
| And please correct the branch name: | Many internally deployed Confluence instances rely on SSO/session login, where API token authentication is not always available in practice. | Added cookie-based authentication support for Confluence API client creation. | 1. Support authenticated export in internal Confluence environments. |
| <font style="color: #ff0000;">That's an em dash (—) in your branch name, not a hyphen.</font> | At the same time, nested tables can lose structure fidelity during Markdown export. | Added dedicated handling for nested tables in Markdown export to keep table structure information. | 1. Preserve nested-table information in exported Markdown. |

UmbraCi:
Hi @Spenhouet! I've successfully separated the nested table fidelity enhancement into this PR (with the real-world export visual comparisons added in the comment above). This cleanly preserves the nested structure as discussed in PR #233. Could you please take a look when you have a moment? Thank you!

Spenhouet:
@UmbraCi Instead of adding a side car file it seems to me that it would be better to directly insert these tables instead in such a case. This does not even seem to need a new config option as the outer table is still displayed as normal markdown table and only the nested table is displayed as HTML, which would otherways never be rendered properly.

UmbraCi:
> @UmbraCi Instead of adding a side car file it seems to me that it would be better to directly insert these tables instead in such a case. This does not even seem to need a new config option as the outer table is still displayed as normal markdown table and only the nested table is displayed as HTML, which would otherways never be rendered properly.

 You mean that if there is a nested table, then create another table outside the same level of the current table, right? If so, how can we ensure that the tables that previously existed in the cells do not lose their location information?

Spenhouet:
Not sure I understand you. I understand the problem this PR solves as follows:

There are nested tables which are currently converted into a nested table of two markdown tables but a markdown table inside a markdown table cell does not render in markdown. Your solution is to detect if it is a nested table (the table inside another table) and in that case to not convert it to a markdown table but instead to keep it as HTML table. 

I suggest that this is directly integrate like that i.e. when a nested table is detected it is keept as HTML instead of converted to markdown.

UmbraCi:
> Not sure I understand you. I understand the problem this PR solves as follows:
> 
> There are nested tables which are currently converted into a nested table of two markdown tables but a markdown table inside a markdown table cell does not render in markdown. Your solution is to detect if it is a nested table (the table inside another table) and in that case to not convert it to a markdown table but instead to keep it as HTML table. 
> 
> I suggest that this is directly integrate like that i.e. when a nested table is detected it is keept as HTML instead of converted to markdown.

When it is detected that the source document to be converted contains nested tables, automatically change the export format from markdown to html.？

UmbraCi:
> Not sure I understand you. I understand the problem this PR solves as follows:
> 
> There are nested tables which are currently converted into a nested table of two markdown tables but a markdown table inside a markdown table cell does not render in markdown. Your solution is to detect if it is a nested table (the table inside another table) and in that case to not convert it to a markdown table but instead to keep it as HTML table. 
> 
> I suggest that this is directly integrate like that i.e. when a nested table is detected it is keept as HTML instead of converted to markdown.

Yes, my way of handling it is to convert nested tables into HTML fragments, because the standard markdown format supports embedded HTML.

Spenhouet:
> When it is detected that the source document to be converted contains nested tables, automatically change the export format from markdown to html.？

Not directly. Only the nested table itself should stay HTML. The rest of the document should be converted as it is right now. 

1. Detect when a table inside a table cell is handled
2. Do not convert it to markdown and leave it as HTML

That's it.

UmbraCi:
Hi @Spenhouet,

Thank you for your valuable feedback! To make sure we are fully aligned on the design:

I need to double check if your expectations are as follows :
1. **Remove all new configurations** (delete `visual_fidelity_markdown` and `visual_fidelity_suffix` entirely).
2. **Remove the sidecar file output** (no additional `*.visual.md` file will be generated; only the main `*.md` file will be exported).
3. **Always preserve nested tables as HTML**: During the normal table conversion process of the main `*.md` file, when a table element is detected nested inside another table cell (e.g. `"table" in parent_tags`), we directly keep the inner table's HTML block (using `str(el)`) inside the outer markdown table cell.

This eliminates any need for user configuration or duplicate files, while naturally supporting high-fidelity nested tables inside GFM/Obsidian renderers.

Is this understanding fully correct? If so, I'll simplify the PR accordingly!

Best regards,

Spenhouet:
Yes, that's what I meant.

UmbraCi:
> Yes, that's what I meant.

done, re-code review @Spenhouet

tests/unit/test_confluence.py
```diff
@@ -175,6 +176,59 @@ def _make_page(
...
+
+        settings = _export_settings(tmp_path)
+        with patch("confluence_markdown_exporter.confluence.settings", settings):
+            page.export_markdown()
+
+        main = tmp_path / "Test Page.md"
+        assert main.exists()
+        assert "<table>" in main.read_text(encoding="utf-8")
+        # Ensure there is no .visual.md or other sidecar
+        assert not (tmp_path / "Test Page.visual.md").exists()
```
Spenhouet: As this was never implemented there is no need to include this check.

UmbraCi: done

confluence_markdown_exporter/utils/table_converter.py
```diff
@@ -67,17 +67,29 @@ class TableConverter(MarkdownConverter):
     """Custom MarkdownConverter for converting HTML tables to markdown tables."""
 
     def convert_table(self, el: BeautifulSoup, text: str, parent_tags: list[str]) -> str:
+        if "table" in parent_tags:
+            return str(el)
+
+        row_tags: list[Tag] = []
+        for child in el.find_all(["thead", "tbody", "tfoot", "tr"], recursive=False):
```
Spenhouet, line 74: What is the purpose of this new code with the row_tags?

Spenhouet (changes requested):
Has one remark and one question

UmbraCi:
The `row_tags` code uses non-recursive traversal (`recursive=False`) to collect table rows and cells.

**Why it is necessary:**
If we use the original recursive `el.find_all("tr")` on the outer table, it will recursively find all `<tr>` elements belonging to both the outer table **and** any nested tables inside its cells. 

As a result, those inner `<tr>` tags would be treated as additional rows of the outer table. This would cause them to be rendered twice: once correctly as part of the raw HTML block inside the cell, and a second time as flattened, corrupted extra rows appended to the end of the outer Markdown table.

By using `recursive=False` (only traversing direct children or direct children of `tbody`/`thead`/`tfoot`), we ensure the converter **only** processes the rows and cells that belong directly to the current table, cleanly isolating its structure from any nested tables within.
@Spenhouet

Spenhouet (approved):

Spenhouet:
Thanks for the explanation and thanks for the implementation. Looks good to me.

UmbraCi:
> Thanks for the explanation and thanks for the implementation. Looks good to me.感谢您的解释和实施，看起来不错。

It looks like there's a Python package that will automatically trigger an update release？

Spenhouet:
The changes were released in version 5.2.0. You might need to wait a moment for pypi to have the package available.

UmbraCi:
Hi @Spenhouet,

I would like to share an in-depth analysis of the bloated trailing whitespace issue in raw Markdown source files when exporting pages containing nested tables, along with some potential solutions for discussion.

#### 1. The Root Cause of Massive Trailing Spaces
Currently, `confluence-markdown-exporter` uses the `tabulate` library to format standard Markdown tables (`tablefmt="pipe"`). When a page contains nested tables, the inner tables are preserved as raw HTML blocks (e.g., `<table><tr><td>...</td></tr></table>`). 

`tabulate` computes column widths by looking at raw string lengths. It has no built-in mechanism to strip or ignore HTML tags (it only strips ANSI escape sequences). Consequently, it treats a 300-character nested HTML string as a 300-character-wide text cell. To maintain plain-text alignment in the raw source, `tabulate` pads every other row in that column with hundreds of trailing spaces to match the 300-character width, resulting in severely bloated Markdown file sizes and poor edit-mode readability.

#### 2. The Plain-Text Visual Alignment Paradox
We explored whether a "plain-text visual alignment" that ignores HTML tags during padding calculation is feasible. However, this is a mathematical paradox:
* **If we pad cells based on raw HTML string lengths**: The raw source code is perfectly aligned, but the rendered preview in Obsidian/GitHub is completely misaligned (because the parsed HTML is rendered much narrower than its raw tag length, while the padded spaces remain).
* **If we pad cells based on visible text lengths (ignoring HTML tags)**: The rendered preview is aligned, but the raw Markdown source in edit mode is still physically misaligned because the HTML string itself physically occupies a large width in the editor.

Therefore, trying to achieve plain-text alignment for cells containing HTML blocks is physically counterproductive.

#### 3. Proposed Solutions for Discussion

We have simulated and tested three potential directions. Here is a comparison of their outputs and trade-offs:

##### **Option A: Custom Compact Table Formatter (Highly Recommended)**
Instead of relying on `tabulate` for padding, we can use a lightweight helper function to output a completely compact, space-free Markdown table.
* **Output Format**:
  ```markdown
  | Col 1 | Col 2 |
  | --- | --- |
  | <table><tr><td>Nested Table</td></tr></table> | Value A |
  | Short text | Value B |
  ```
* **Pros**: 
  * Extreme optimization: reduces raw Markdown file size by **70%+** for pages with large tables.
  * Perfect edit-mode readability: no massive trails of spaces.
  * Rendered preview is **100% correct**: engines like Obsidian and GitHub GFM auto-align the columns perfectly during rendering anyway.
  * Extremely easy to implement (roughly 8 lines of code) and eliminates the need to bend `tabulate` to fit this niche use-case.
* **Cons**: Breaks visual alignment inside the raw/edit mode (which, as shown above, is already broken/impossible for HTML cells).

##### **Option B: Tabulate with Padding Disabled (`stralign=None`, `numalign=None`)**
We can pass `stralign=None` and `numalign=None` to `tabulate`.
* **Output Format**:
  ```markdown
  | Col 1 | Col 2 |
  |----------------------------------------------------------------------|---------|
  | <table><tr><td>Nested Table</td></tr></table> | Value A |
  | Short text | Value B |
  ```
* **Pros**: No trailing spaces in data rows.
* **Cons**: The header separator line under the header is still calculated using maximum raw width, resulting in an extremely long, unbalanced line of dashes (`---...---`), which still bloats the file and looks very messy in raw text.

---

### Seeking Your Feedback
Since `tabulate` is only called in two specific files (`table_converter.py` and `confluence.py` for meta-bind-view-fields), replacing it with the **Compact Formatter (Option A)** seems to be the most robust, lightweight, and standard approach for Markdown exporters.

Which solution do you think is more reasonable and aligned with the project's vision? I look forward to your thoughts before proceeding with the implementation and updating our tests!

Best regards,

Spenhouet:
I suggest we could add a config option `table_column_width_alignment` which is true by default. If true it uses tabulate to format the tables, if false it does not and outputs the compact table.

UmbraCi:
> I suggest we could add a config option `table_column_width_alignment` which is true by default. If true it uses tabulate to format the tables, if false it does not and outputs the compact table.

If you have nested tables, wouldn't it be better to skip the 'tabulate' function and output a compact table? That is, table_column_width_alignment defaults to false in the case of nested tables

Spenhouet:
We could make the config table_column_width with the options "aligned", "mixed" and "compact".

Where mixed only aligned the width if the table stays below a certain width?

UmbraCi:
> We could make the config table_column_width with the options "aligned", "mixed" and "compact".
> 
> Where mixed only aligned the width if the table stays below a certain width?

We still include the table_column_width_alignment configuration (true/false)
When configured to the default value of true, the smart "mixed" mode is automatically enabled under the surface:
Fully intelligent conversion: When exported, the tool automatically detects if the table is nested (with nested <table>) or if the total aligned width exceeds 120 characters.
If so, it is automatically exported in compact format. If it's just a small table, you can still keep the nice alignment.
Out of the box, the average user gets the best results without any extra configuration!
When you explicitly set it to false, it forces all tables to be exported in compact format.

Spenhouet:
The new table_column_width string/literal based config replaces the suggested table_column_width_alignment config. Only the table_column_width config would be implemented

UmbraCi:
Hi @Spenhouet,

Following our discussion, we have implemented the proposed `table_column_width` configuration option which supports three modes: `"aligned"`, `"compact"`, and `"mixed"` (Default).

We ran test exports of a page containing a nested table under all three modes so you can see the actual visual effects and compare them directly:

### 1. Aligned Mode (`table_column_width = "aligned"`)
* **Behavior**: Always align table columns using spaces (legacy behavior).
* **Actual Output**:
```markdown
# Nested Table Export Comparison

| 1                                                                                                                                                                                                                                                                                                                                                                                          | 2                                                                                                                                            | 3                                                                                                  | 4                                                                    |
| :----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | :------------------------------------------------------------------------------------------------------------------------------------------- | :------------------------------------------------------------------------------------------------- | :------------------------------------------------------------------- |
| <table class="confluenceTable"><tbody><tr><th class="confluenceTh">1.1</th><th class="confluenceTh">1.2</th><th class="confluenceTh">1.3</th></tr><tr><td class="confluenceTd">1.4</td><td class="confluenceTd">1.5</td><td class="confluenceTd">1.6</td></tr><tr><td class="confluenceTd">1.7</td><td class="confluenceTd">1.8</td><td class="confluenceTd">1.9</td></tr></tbody></table> | Background                                                                                                                                   | Goal                                                                                               | Changes                                                              |
| And please correct the branch name:                                                                                                                                                                                                                                                                                                                                                        | Many internally deployed Confluence instances rely on SSO/session login, where API token authentication is not always available in practice. | Added cookie-based authentication support for Confluence API client creation.                      | 1. Support authenticated export in internal Confluence environments. |
| <font style="color: #ff0000;">That's an em dash (—) in your branch name, not a hyphen.</font>                                                                                                                                                                                                                                                                                              | At the same time, nested tables can lose structure fidelity during Markdown export.                                                          | Added dedicated handling for nested tables in Markdown export to keep table structure information. | 1. Preserve nested-table information in exported Markdown.           |
```
* **Review**: In this mode, the raw HTML in the first cell of the first row causes `tabulate` to pad the entire column with hundreds of spaces, making the plain-text editing view extremely wide and difficult to read.

---

### 2. Compact Mode (`table_column_width = "compact"`)
* **Behavior**: Never pad with spaces, yielding minimal and clean GFM pipe tables.
* **Actual Output**:
```markdown
# Nested Table Export Test

| 1 | 2 | 3 | 4 |
| --- | --- | --- | --- |
| <table class="confluenceTable"><tbody><tr><th class="confluenceTh">1.1</th><th class="confluenceTh">1.2</th><th class="confluenceTh">1.3</th></tr><tr><td class="confluenceTd">1.4</td><td class="confluenceTd">1.5</td><td class="confluenceTd">1.6</td></tr><tr><td class="confluenceTd">1.7</td><td class="confluenceTd">1.8</td><td class="confluenceTd">1.9</td></tr></tbody></table> | Background | Goal | Changes |
| And please correct the branch name: | Many internally deployed Confluence instances rely on SSO/session login, where API token authentication is not always available in practice. | Added cookie-based authentication support for Confluence API client creation. | 1. Support authenticated export in internal Confluence environments. |
| <font style="color: #ff0000;">That's an em dash (—) in your branch name, not a hyphen.</font> | At the same time, nested tables can lose structure fidelity during Markdown export. | Added dedicated handling for nested tables in Markdown export to keep table structure information. | 1. Preserve nested-table information in exported Markdown. |
```
* **Review**: Very tight and clean in the plain-text editor, but lacks column alignment visual comfort for simple tables.

---

### 3. Mixed Mode (Default, `table_column_width = "mixed"`)
* **Behavior**: Smart hybrid mode. Keeps simple/narrow tables aligned using spaces, but automatically falls back to compact mode if a cell contains a nested `<table>` tag OR the aligned table line length exceeds `120` characters.
* **Actual Output**: (Same as Compact above because it contains a nested table and auto-downgraded to compact).
* **Review**: This provides the best of both worlds! Simple metadata tables remain aligned and beautiful to edit, while complex/nested tables are compressed compactly to prevent space-padding bloat.

---

### Proposing an Additional Optimization: HTML Attribute & Newline Cleaning

During our testing, we noticed that the raw HTML nested table in the first cell still retains Confluence-specific attributes (like `class="confluenceTable"`, `class="confluenceTh"`, `class="confluenceTd"`) and can sometimes contain redundant newlines that risk breaking the outer pipe table structure.

We would like to propose a small additional optimization: **we can clean up these Confluence-specific HTML attributes and redundant newlines**, leaving only standard clean HTML tags (such as `<table>`, `<tr>`, `<td>`, preserving only essential layout attributes like `colspan`, `rowspan`, and `align`).

With this optimization, the nested table cell content would become extremely clean:
```html
<table><tbody><tr><th>1.1</th><th>1.2</th><th>1.3</th></tr><tr><td>1.4</td><td>1.5</td><td>1.6</td></tr><tr><td>1.7</td><td>1.8</td><td>1.9</td></tr></tbody></table>
```

This keeps the visual rendering in Markdown previewers identical while reducing plain-text editing clutter by over 60%!

Please let us know what you think of the three modes and the additional cleaning proposal. If you agree, we will refine the implementation and push the branch to open a new PR.

PR #277, opened 2026-07-10
fix: fetch all Page Properties Report rows instead of first page

### LINKED ISSUE #276: Page Properties Report macro exports only the first page of items (configured page size)
### Description

When a page contains a Page Properties Report macro (`detailssummary`) with "Number of items to display" set to e.g. 30, the exported Markdown table contains only those first 30 rows instead of all matching items.

Reproduce: create a Page Properties Report whose CQL matches more items than the configured page size, export the page (e.g. `cme pages <page-url>`) with the default `export.page_properties_report_format = "frozen"`, and compare row counts.

### Cause

`convert_page_properties_report` (`confluence_markdown_exporter/confluence.py`, lines 2701-2717) converts the table that Confluence pre-rendered into `body.export_view`, which Confluence truncates to the macro's page size. The `data-cql` attribute on the table is never executed against the API, so the remaining rows are never fetched.

### Proposed fix

In `frozen` mode, fetch all rows via `rest/masterdetail/1.0/detailssummary/lines` (the same endpoint the macro's own pagination UI uses), looping over `pageIndex` until `totalPages` is reached, and build the full table from the returned server-rendered rows. Fall back to the current truncated snapshot on any error, so behavior never regresses.

I can submit a PR for this.

### Version

confluence-markdown-exporter 5.2.1 (current main)

### Confluence Version

Confluence Cloud (Data Center presumably affected as well)

stn1slv:
## Summary

Fixes #276.

When a page contains a Page Properties Report macro (`detailssummary`) with "Number of items to display" set to e.g. 30, the exported Markdown table contained only those first 30 rows instead of all matching items. The `frozen` export copied the table pre-rendered into `body.export_view`, which Confluence truncates to the macro's configured page size; the `data-cql` attribute was never executed against the API.

This change makes the `frozen` format fetch **all** report rows through `rest/masterdetail/1.0/detailssummary/lines` (the same REST endpoint the macro's own pagination UI uses), looping over `pageIndex` until `totalPages` is exhausted, and rebuilds one complete table:

- New `_fetch_page_properties_report_table` pages through the endpoint (50 rows per request) using the parameters already present on the rendered table (`data-cql`, `data-current-space-key`, `data-headings`, `data-sort-by`, `data-reverse-sort`), so the report's configured sort order is preserved.
- New `_build_report_table` assembles the full HTML table. The first column is emitted as a proper page link (`data-linked-resource-type="page"` + `data-linked-resource-id`), so links resolve to exported `.md` paths exactly like native page links. Detail cells reuse the server-rendered HTML fragments, so no property-extraction logic is duplicated. Rows whose detail line has no numeric id keep the plain title text instead of a link.
- **Graceful degradation — the output can never get worse than before.** The export falls back to the previous behavior (the rendered snapshot) whenever the full row set cannot be trusted:
  - HTTP error from the endpoint (e.g. instances without it);
  - unexpected/malformed response shape;
  - empty result;
  - fewer rows collected than the response's reported `total` (a partial table that looks complete would silently drop rows);
  - the macro has no explicit column configuration (empty `data-headings`), where a rebuilt header would not match the server-derived detail cells.
- The `dataview` format is unchanged. Setting description and `docs/configuration/options.md` are updated accordingly.

## Test Plan

- New unit test class `TestPagePropertiesReportFrozenFetchesAllRows`:
  - rows merged across two API result pages (asserts `pageIndex` 0 and 1 are requested and all rows plus headings appear in the Markdown table);
  - single-page report performs exactly one request;
  - empty `detailLines` falls back to the rendered snapshot;
  - malformed response (non-dict detail line) falls back to the snapshot;
  - a detail line with `"details": null` keeps the row with the title link;
  - a detail line without a numeric id renders plain title text (no crash);
  - an empty later page falls back to the snapshot instead of exporting partial rows;
  - an HTTP error on the second page request discards partial rows and falls back;
  - empty `data-headings` skips the fetch entirely and keeps the snapshot.
- Updated `test_frozen_table_falls_back_to_snapshot_when_endpoint_unavailable`: an `HTTPError` from the endpoint falls back to the rendered snapshot.
- Full suite: 442 passed; `ruff check` clean.

### End-to-end verification on Confluence Cloud

Verified against a real Confluence Cloud instance with a Page Properties Report configured to 30 items per page and 35 matching pages:

- Before the fix, `body.export_view` contained a 30-row snapshot table (confirmed in the debug HTML dump); the export was truncated accordingly.
- With the fix, the exported Markdown table contains **all 35 rows**, in the report's configured order, with the first column resolving to the exported `.md` files. No fallback warnings were logged.
- Multi-page fetching was verified live against the same instance by calling the endpoint with a small page size (10): iterating `pageIndex` over the reported 4 `totalPages` collected all 35 rows with no duplicates or gaps across page boundaries — the same loop logic the production code uses.
- Cloud specifics confirmed: the export view renders the report with the same `metadata-summary-macro` / `data-cql` markup as Server/DC, and `rest/masterdetail/1.0/detailssummary/lines` is available on Cloud (served via the cloud gateway). The Cloud response omits the `total` field, which the completeness check tolerates (`response.get("total") or 0`).
- Data Center was not tested live; the endpoint and table markup originate from Server/DC, and instances without the endpoint keep today's behavior via the fallback.

### DIFF
--- a/confluence_markdown_exporter/confluence.py
+++ b/confluence_markdown_exporter/confluence.py
@@ -4,6 +4,7 @@ https://developer.atlassian.com/cloud/confluence/rest/v1/intro
 """
 
 import functools
+import html
 import json
 import logging
 import mimetypes
@@ -77,6 +78,9 @@ StrPath: TypeAlias = str | PathLike[str]
 
 logger = logging.getLogger(__name__)
 _MAX_UNICODE_CODEPOINT = 0x10FFFF
+
+# Rows requested per call when fetching all Page Properties Report entries.
+_REPORT_FETCH_PAGE_SIZE = 50
 
 _RE_RGB_BG = re.compile(r"background-color:\s*rgb\((\d+),\s*(\d+),\s*(\d+)\)")
 _RE_RGB_COLOR = re.compile(r"(?<![a-z-])color:\s*rgb\((\d+),\s*(\d+),\s*(\d+)\)")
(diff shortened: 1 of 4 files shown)


stn1slv:

Spenhouet (approved):

Spenhouet:
@stn1slv Thanks for the addition

PR #282, opened 2026-07-20
fix: export page properties report nested in third-party app macros

### LINKED ISSUE #281: Page Properties Report is not exported when nested inside a third-party app macro (StiltSoft Table Filter)
### Description

A Page Properties Report (`detailssummary`) that exports correctly on its own produces no table at all once it is nested inside a third-party Connect app macro, such as the StiltSoft **Table Filter** macro (`table-filter`). No warning is logged and no placeholder is emitted — the table is silently absent from the exported Markdown.

Reproduce:

1. Create a page with a Page Properties Report macro and export it, e.g. `cme pages https://company.atlassian.net/wiki/spaces/KEY/pages/123/Title`. The table is exported correctly.
2. Edit the page and wrap that same report in a Table Filter macro (the report becomes the Table Filter's body).
3. Export again with the same command. The table is now missing entirely from the output.

Tested with the default `export.page_properties_report_format = "frozen"`.

### Cause

Confluence does not server-render a Connect `dynamicContentMacros` app macro's body into `body.view`. It emits only an empty placeholder plus an iframe bootstrap script:

```html
<div class="ap-container conf-macro output-block"
     data-macro-name="table-filter" data-macro-id="..." data-hasbody="true">
  <div class="ap-content"> </div>
  <script class="ap-iframe-body-script">...</script>
</div>
```

The nested report appears in `body.view` only as a truncated string inside that script's JSON options (`"macro.body": "<div ... data-macro-name=\"detailssummary\" ...`, alongside `"macro.truncated": "true"`), so it cannot be recovered from there.

Since the converter operates on `body.view` (`Page.html` → `Converter.markdown`), the `<table class="metadata-summary-macro">` element never exists in the input. `convert_table` (`confluence_markdown_exporter/confluence.py:2699`) therefore never dispatches to `convert_page_properties_report`, and nothing is emitted. This is why the failure is silent: the existing `return ""` guards inside `convert_page_properties_report` are never even reached.

`body.export_view`, by contrast, renders the report **fully**, with the app macro wrapper stripped and every attribute the existing code path needs still intact: `data-cql`, `data-headings`, `data-sort-by`, `data-reverse-sort` and `data-current-space-key`.

### Proposed fix

As a preprocessing step before conversion, replace each empty `div.ap-container[data-hasbody="true"]` in the view HTML with the corresponding report table taken from `body.export_view`.

The two must be joined via the `cql` parameter of the nested `detailssummary` macro in `body.storage`, because `body.export_view` carries no `data-macro-id` or `data-macro-name` attributes at all, leaving the CQL string as the only reliable key.

Existing handling then applies unchanged, including the full-pagination fetch added in #277 — which matters here, since the `export_view` snapshot is still limited to the macro's configured page size.

Keying on `ap-container` rather than on the literal macro name `table-filter` means this also covers other third-party wrappers (Table Excerpt, Chart from Table, and similar) at no extra cost.

Known limitation worth stating up front: Table Filter applies its filtering, sorting and column hiding client-side inside the app iframe, and none of that is represented in the REST data. The export is therefore the full, unfiltered report rather than the filtered view seen in the browser.

I can submit a PR for this.

### Version

confluence-markdown-exporter 5.3.0 (current main)

### Confluence Version

Confluence Cloud

---

**Update (during implementation, see #282):** one non-obvious detail worth recording here. `body.storage` must be parsed with `html.parser`, **not** the `xml` parser used for the other storage lookups in `confluence.py`.

Confluence storage can contain HTML entities that are undefined in XML — third-party macro parameters emit e.g. `&sbquo;`. Those put lxml into error recovery, where it silently drops *unrelated* entities elsewhere in the same document, including the `&quot;` wrapping CQL values. The CQL then reads `label = my-label and ...` instead of `label = "my-label" and ...`, so the join key never matches and the report is dropped anyway.

The same fragility applies in principle to the existing `xml`-parser storage lookups (`_storage_plantuml_macros`, `_extract_include_target_title`, `_extract_uml_from_editor2`). Those are untouched in #282 to keep it scoped; happy to open a separate issue if you would like them hardened.

stn1slv:
## Summary

Closes #281.

A Page Properties Report that exports correctly on its own produced **no table at all** once it was nested inside a third-party Connect app macro such as StiltSoft Table Filter — silently, with no warning.

Confluence does not server-render the body of a Connect `dynamicContentMacros` app macro into `body.view`. It emits only an empty placeholder plus an iframe bootstrap script:

```html
<div class="ap-container conf-macro output-block"
     data-macro-name="table-filter" data-macro-id="..." data-hasbody="true">
  <div class="ap-content"> </div>
  <script class="ap-iframe-body-script">...</script>
</div>
```

The nested report survives there only as a **truncated** string inside that script's JSON options (alongside `"macro.truncated": "true"`), so it cannot be recovered from `body.view`. The `<table class="metadata-summary-macro">` element simply never exists in the converter's input, so `convert_table` never dispatches to `convert_page_properties_report`. Nothing was dropping the table; it was never there. That is also why the failure was silent — the existing `return ""` guards were never reached.

`body.export_view` does render the report fully, with the app macro wrapper stripped and every attribute the existing code path needs still intact (`data-cql`, `data-headings`, `data-sort-by`, `data-current-space-key`).

So this is a **routing problem, not a conversion problem**. The fix is one preprocessing pass, `_inline_app_macro_bodies`, added alongside the existing `_strip_excerpt_include_panel_titles` pass in `Converter.markdown`. It replaces each empty `ap-container` placeholder with the matching report table from `body.export_view`, after which all existing handling applies unchanged — including the full-pagination fetch from #277, and the `dataview` output format.

Matching is done on the CQL of the nested `detailssummary` macro in `body.storage`, because `body.export_view` contains no `data-macro-id` or `data-macro-name` attributes at all, leaving the CQL as the only reliable join key.

Keying on `ap-container` rather than on the literal name `table-filter` means other third-party wrappers (Table Excerpt, Chart from Table, and similar) are covered at no extra cost.

### A parser detail worth flagging

`_nested_report_cqls` parses `body.storage` with `html.parser` rather than the `xml` parser used elsewhere in this file.

Confluence storage can contain HTML entities that are **undefined in XML** — third-party macro parameters emit e.g. `&sbquo;`. Those put lxml into error recovery, where it silently drops *unrelated* entities elsewhere in the same document, including the `&quot;` wrapping CQL values. The result is a CQL like `label = my-label and ...` instead of `label = "my-label" and ...`, so the join key never matches.

This cost me a debugging cycle, and it is worth being aware of: the same fragility applies in principle to the other `xml`-parser storage lookups in this file (`_storage_plantuml_macros`, `_extract_include_target_title`, `_extract_uml_from_editor2`). I have deliberately **not** changed those here to keep this PR scoped, but I am happy to open a follow-up issue if you would like them hardened too.

### Logging behaviour

Because the original bug's defining trait was that content vanished with **nothing logged**, the failure paths here are explicit:

- A report present in `body.storage` but with no matching table in `body.export_view` logs a **warning** and is omitted.
- Multiple reports sharing an identical CQL log a **warning** (only the first rendered table can be indexed, since `export_view` has no macro IDs).
- An app macro that wraps no report at all logs at **debug** only. Most app macros legitimately have no report inside them, so warning here would be pure noise.

The CQL string itself is deliberately kept out of warning-level messages and logged at debug instead. It is macro configuration and can embed labels or free-text filters, and every other warning in this module logs metadata only, never a macro's parameter values.

### Known limitations

- Table Filter applies its filtering, sorting and column hiding **client-side** inside the app iframe, and none of that is represented in the REST data. The export is therefore the full, unfiltered report rather than the filtered view seen in the browser. This is documented in `docs/features.md`.
- Static content (for example a plain `<table>`) inside an app macro is still not recovered, since there is no join key available for it. Out of scope here.

## Test Plan

`uv run pytest` — **459 passed** (442 before this change, 17 new). `uv run ruff check` clean.

New test class `TestAppMacroNestedPagePropertiesReport` covers:

- placeholder replaced with the report table, and the report reaching the Markdown output
- the pass being wired into `Converter.markdown`, not merely available
- HTML without an app macro returned unchanged
- an app macro that **did** render content server-side being left alone, so nothing is discarded
- unknown macro id, CQL absent from `body.export_view`, and missing `body.storage` all degrading quietly instead of crashing
- two reports inside one wrapper, both emitted in document order
- the same report referenced by two placeholders being emitted twice
- nested `ap-container` placeholders not crashing the export
- the logging contract above: warning on an unresolvable report, debug-only for an app macro with no report, and no raw CQL at warning level
- `export.page_properties_report_format = "dataview"` still applying to a spliced report
- **regression test for the entity issue above** — storage containing `&sbquo;` alongside `&quot;`-wrapped CQL

The five tests that pin a specific defect — the entity corruption, the shared-table copy, the nested-placeholder crash, and the two logging assertions — were each checked by reverting the corresponding fix and confirming the test fails, so they do not pass vacuously.

Fixtures include the iframe bootstrap `<script>` that the real placeholder always carries. An earlier fixture that omitted it passed while the real page still failed, which is worth mentioning because it is the trap this whole area invites.

Verified end-to-end against a real Confluence Cloud page with a Page Properties Report nested in a Table Filter macro: the table is absent before the change and fully exported after it, with **35 data rows** — more than the 30-row `body.export_view` snapshot, confirming the pagination path runs rather than the truncated snapshot. No leftover placeholder markup, no warnings emitted.

### Review history

This branch went through two review passes. The first found two real defects, both since fixed and covered by the tests above:

- `insert_before` **moves** a node rather than copying it, so a table referenced by two placeholders was silently lost — now inserts a copy.
- Placeholders are collected before the mutation loop, so a nested `ap-container` crashed the whole export once the outer one was decomposed — now guarded.

The second pass found no defects; it produced the logging refinements described above.

### DIFF
--- a/confluence_markdown_exporter/confluence.py
+++ b/confluence_markdown_exporter/confluence.py
@@ -3,6 +3,7 @@
 https://developer.atlassian.com/cloud/confluence/rest/v1/intro
 """
 
+import copy
 import functools
 import html
 import json
@@ -1501,6 +1502,7 @@ class Page(Document):
             self._panel_icon_map_cache: dict[str, str] | None = None
             self._plantuml_index: int = 0
             self._storage_plantuml_macros_cache: list[Tag] | None = None
+            self._storage_soup_cache: BeautifulSoup | None = None
 
         @property
         def _colorid_map(self) -> dict[str, str]:
@@ -2607,6 +2609,10 @@ class Page(Document):
                 return html
 
             soup = BeautifulSoup(html, "html.parser")
+            # `get_text` ignores <script>/<style> contents (BeautifulSoup models them as
+            # Script/Stylesheet, which are excluded from the string traversal), so the
+            # iframe bootstrap script the placeholder always carries does not count as
+            # rendered content. Only a macro the app did not render server-side matches.
             placeholders = [
                 el
                 for el in soup.find_all("div", class_="ap-container")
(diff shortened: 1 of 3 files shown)


stn1slv:

Spenhouet (approved):

PR #284, opened 2026-07-21
fix: render content-tree (page-tree) macro as nested page links

### LINKED ISSUE #283: Content Tree (Page Tree) macro is dropped from export
## Description

The **Content Tree** macro (also shown as "Page Tree"; storage name `content-tree` / legacy `pagetree`) is silently dropped during export. Pages that use it as a navigation index export with an empty body: only the page title is written, and the tree of child-page links is lost.

## Why it happens

The exporter converts the rendered `body.view` HTML with `markdownify`, dispatching macros by their `data-macro-name` in `Page.Converter.convert_div`. There is no handler for the content-tree / pagetree macro and no generic fallback, so it falls through to the base `convert_div`. In `body.view` this macro is an empty JavaScript placeholder (its tree is loaded via AJAX), so nothing is emitted.

## Steps to reproduce

1. Create a Confluence page with child pages.
2. Add the **Content Tree** (Page Tree) macro to the page (e.g. Root content `@self`).
3. Export the page to Markdown.

**Expected:** the exported markdown contains a nested list of links to the child pages.
**Actual:** the macro produces no output; the page body is empty apart from the title.

## Proposed fix

Add a `convert_pagetree` handler that resolves the tree's child pages via the API (reusing `Page.descendants` and `convert_page_link`) and emits a nested Markdown bullet list of page links, honoring the macro's `root` (`@self` / `@home` / page title) and tree-depth parameters. Detect the macro by `data-macro-name` (`content-tree`, `pagetree`, `children`) and by CSS class (`content-tree`, `plugin_pagetree`) as a fallback.

I have a working implementation and will open a PR referencing this issue.

stn1slv:
## What

Adds a handler for the **Content Tree** / **Page Tree** macro (`content-tree`, legacy `pagetree`, and the related `children` / Children Display macro). It resolves the tree's child pages from the API and emits a nested Markdown bullet list of page links.

## Why

The macro renders as an empty JavaScript placeholder in `body.view`, so it was silently dropped during export. Pages that use it as a navigation index exported with an empty body (only the title). See #283.

## How

- New `convert_pagetree` handler, dispatched from `convert_div` by `data-macro-name` (`content-tree`, `pagetree`, `children`) and by CSS class (`content-tree`, `plugin_pagetree`) as a fallback for when the rendered placeholder lacks `data-macro-name`.
- Macro parameters are read from the storage-format XML (`editor2` / `body_storage`), matched by `macro-id` (mirroring the existing `include` macro handling). The `root` value is taken from its nested `ri:page` `ri:content-title` attribute, which is how Confluence actually stores it (verified against a live page), with a fallback to element text for other variants.
- Root resolution: `@self` / empty -> current page; `@home` -> space homepage; a page title -> best-effort CQL lookup. The lookup honors the root reference's `ri:space-key` (so a root in another space resolves correctly), falls back to the current space when absent, escapes backslashes/quotes, and requires an exact title match so a fuzzy CQL hit cannot resolve to the wrong page.
- Depth: `depth` (Children Display) is preferred, then `startDepth` (Page Tree, which maps to the same "Tree depth" UI field on the instances observed). The `children` macro defaults to direct children only unless `all=true` or an explicit depth is set; Content Tree / Page Tree default to the full subtree.
- The tree is built from `Page.descendants`, nested via each node's ancestor chain and limited to the configured depth. Children are ordered by title (Confluence's manual "position" order is not available via the CQL descendants query). Descendants with a missing id are skipped so one malformed node cannot abort the page.
- Links reuse the existing `convert_page_link` helper, so wiki/relative href modes and title disambiguation are handled consistently.

## Tests

New `tests/unit/test_pagetree_conversion.py` covering: nested rendering, depth limiting, `@self` / `@home` roots, missing homepage, unresolvable title root, empty subtree, dispatch via `convert_div`, plain-text vs. `ri:content-title` root, cross-space root resolution, exact-title rejection, the `children` default/`all`/`depth` behavior, the unmatched-`macro-id` warning, and the zero-id descendant guard. Full suite passes (457 tests); lint and format clean.

## Notes

- Verified end-to-end against a live Confluence page: the previously empty page now renders its 15 child pages as links.
- The exact rendered markup (`data-macro-name` vs. CSS class) can vary between Confluence Cloud and Server/DC; detection covers both.

Closes #283.

### DIFF
--- a/confluence_markdown_exporter/confluence.py
+++ b/confluence_markdown_exporter/confluence.py
@@ -1984,6 +1984,10 @@ class Page(Document):
                 return ""
 
             max_depth = self._pagetree_depth(params)
+            if max_depth is None and str(el.get("data-macro-name", "")) == "children":
+                # The Children Display macro shows only direct children unless `all=true`.
+                if params.get("all", "").strip().lower() != "true":
+                    max_depth = 1
             lines = self._render_pagetree(root_page.id, root_page.descendants, max_depth)
             if not lines:
                 return ""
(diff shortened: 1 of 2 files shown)


stn1slv:

Spenhouet (approved):

PR #289, opened 2026-08-04
feat(confluence): export plantumlcloud macro as a code block

### LINKED ISSUE #290: PlantUML diagrams from the Cloud `plantumlcloud` macro are silently dropped from the export
### Description

Pages that contain diagrams from the [Flowchart, PlantUML Diagrams for Confluence](https://marketplace.atlassian.com/apps/1222993/flowchart-plantuml-diagrams-for-confluence) app export with the diagram **missing entirely**: no code block, no image, and no marker of any kind. The surrounding text is exported normally, so the omission is easy to miss.

`docs/features.md` links to this exact marketplace app and states that PlantUML is "exported as fenced PlantUML code blocks", so the documented behaviour and the actual behaviour differ.

**Command:**

```
cme pages https://company.atlassian.net/wiki/spaces/KEY/pages/123/Title
```

**Expected:** a fenced ```` ```plantuml ```` block per diagram, as already happens for the Server/Data Center `plantuml` macro.

**Actual:** nothing is emitted where each diagram was.

### Cause

On Confluence **Cloud**, this app registers the macro name **`plantumlcloud`**, not `plantuml`. The `macro_handlers` map in `Page.Converter.convert_div` (`confluence_markdown_exporter/confluence.py`) dispatches `plantuml` only, so `plantumlcloud` falls through to `super().convert_div(...)`.

It is a Connect (iframe) app, so `body.view` contains only an empty placeholder:

```html
<div class="ap-container conf-macro output-block" data-macro-name="plantumlcloud" data-macro-id="...">
  <div class="ap-content"> </div>
  <script>...</script>
</div>
```

markdownify drops `<script>`, and `ap-content` is empty, so the fall-through converts the children to an empty string. The app also writes nothing usable to `body.editor2`.

The diagram source is present, but only in `body.storage`, in the macro's `data` parameter:

```xml
<ac:structured-macro ac:name="plantumlcloud" ac:macro-id="...">
  <ac:parameter ac:name="filename">diagram.svg</ac:parameter>
  <ac:parameter ac:name="data">...</ac:parameter>
  <ac:parameter ac:name="compressed">true</ac:parameter>
</ac:structured-macro>
```

The `data` value decodes with the mxgraph/draw.io pipeline: **base64 decode, raw DEFLATE inflate (`wbits = -15`), percent-decode**. When the sibling `compressed` parameter is absent or not `true`, the deflate step is skipped. Confirmed against a page holding 5 such diagrams; all 5 decoded to plain `@startuml ... @enduml` text.

### Notes

Because the macro is unhandled rather than explicitly skipped, there is no warning and no placeholder comment, so an export can lose every diagram on a page without any signal to the user.

A fix is proposed in #289.

### Version

confluence-markdown-exporter 5.3.0

### Confluence Version

Confluence Cloud

stn1slv:
Fixes #290

## Summary

Pages using the [Flowchart, PlantUML Diagrams for Confluence](https://marketplace.atlassian.com/apps/1222993/flowchart-plantuml-diagrams-for-confluence) app currently export with the diagram missing entirely: no code block, no image, not even a marker.

On Confluence Cloud that app registers the macro name **`plantumlcloud`**, while only `plantuml` was dispatched in `macro_handlers`. Being a Connect app, it renders nothing but an empty iframe placeholder into `body.view` and writes nothing to `editor2`, so the unhandled macro fell through to markdownify's default `convert_div` and produced an empty string.

`docs/features.md` already linked to this exact marketplace app and promised fenced PlantUML blocks, so this closes a gap between the documented and actual behaviour.

### What it does

The diagram source lives in `body.storage`, in the macro's `data` parameter: base64, and when the sibling `compressed` parameter is `true`, additionally raw-DEFLATE-compressed and percent-encoded (the mxgraph/draw.io encoding). The new `convert_plantumlcloud` handler recovers it and emits a fenced `plantuml` block, matching what the existing Server/DC `plantuml` handler produces.

The rendered `.svg`/`.png` attachments are deliberately **not** embedded; the diagram source is exported, as documented.

### Decoding is deliberately defensive

`body.storage` is remote content, so every failure mode degrades to a visible `<!-- PlantUML diagram (source not found) -->` marker plus a log warning, never to wrong or empty output:

- the inflate is bounded at 4 MB and rejects truncated streams, which inflate without raising and would otherwise be exported as a silently partial diagram;
- base64 decoding tolerates line wrapping, but never falls back to emitting the undecoded parameter;
- percent-decoding is applied only when the text is actually encoded, because PlantUML sources legitimately contain a literal `%` (`%date%`, `%filename()`).

### Macro matching

Macros are matched by `macro-id`, falling back to document order only when `body.storage` carries no `macro-id` at all. A placeholder whose `macro-id` is unknown to this page belongs to a transcluded page (an `include` macro expands another page's view HTML inline), so it resolves to the marker rather than to an unrelated diagram.

### Refactors carried along

- `_storage_plantuml_macros` generalised into `_storage_macros_by_name(name)` with a name-keyed cache; the existing `plantuml` path is unchanged.
- The parameter-extraction loop duplicated between the panel-icon code and the new code extracted into `_macro_params()`.

## Test Plan

- `tests/unit/test_plantumlcloud_conversion.py` adds 24 tests using synthetic diagram sources: happy path, compressed and uncompressed payloads, line-wrapped base64, literal `%` preservation, `macro-id` selection, positional fallback and cursor semantics, foreign/transcluded `macro-id`, truncated and oversized payloads, nested parameters, and a case with both macro flavours on one page guarding the shared cache.
- `uv run pytest`: **466 passed** (442 before this change).
- `uv run ruff check` and `uv run ruff format --check`: clean.
- `uv build --no-sources`: builds.
- Verified end to end against a real Cloud page holding 5 `plantumlcloud` diagrams: all 5 export as `plantuml` fences with no "source not found" markers.

### DIFF
--- a/confluence_markdown_exporter/confluence.py
+++ b/confluence_markdown_exporter/confluence.py
@@ -3,6 +3,7 @@
 https://developer.atlassian.com/cloud/confluence/rest/v1/intro
 """
 
+import base64
 import functools
 import html
 import json
@@ -11,6 +12,7 @@ import mimetypes
 import os
 import re
 import urllib.parse
+import zlib
 from collections.abc import Set
 from concurrent.futures import ThreadPoolExecutor
 from concurrent.futures import as_completed
@@ -81,6 +83,10 @@ _MAX_UNICODE_CODEPOINT = 0x10FFFF
 
 # Rows requested per call when fetching all Page Properties Report entries.
 _REPORT_FETCH_PAGE_SIZE = 50
+
+# Upper bound for an inflated `plantumlcloud` payload. Deflate reaches ~1000:1, so an
+# unbounded inflate of remote page content could exhaust memory.
+_MAX_PLANTUML_SOURCE_BYTES = 4 * 1024 * 1024
 
 _RE_RGB_BG = re.compile(r"background-color:\s*rgb\((\d+),\s*(\d+),\s*(\d+)\)")
 _RE_RGB_COLOR = re.compile(r"(?<![a-z-])color:\s*rgb\((\d+),\s*(\d+),\s*(\d+)\)")
@@ -1478,7 +1484,8 @@ class Page(Document):
             self._image_captions_cache: dict[str, str] | None = None
             self._panel_icon_map_cache: dict[str, str] | None = None
             self._plantuml_index: int = 0
-            self._storage_plantuml_macros_cache: list[Tag] | None = None
+            self._plantumlcloud_index: int = 0
+            self._storage_macros_cache: dict[str, list[Tag]] = {}
 
         @property
         def _colorid_map(self) -> dict[str, str]:
@@ -1494,10 +1501,9 @@ class Page(Document):
                 self._colorid_map_cache = cache
             return self._colorid_map_cache
 
-        @property
-        def _storage_plantuml_macros(self) -> list[Tag]:
-            """Cache and return all PlantUML structured-macros from body.storage."""
-            if self._storage_plantuml_macros_cache is None:
+        def _storage_macros_by_name(self, name: str) -> list[Tag]:
+            """Cache and return all structured-macros with the given name from body.storage."""
+            if name not in self._storage_macros_cache:
                 macros: list[Tag] = []
                 if self.page.body_storage:
                     wrapped = f"<root>{self.page.body_storage}</root>"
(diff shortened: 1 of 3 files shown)


stn1slv:

Spenhouet (approved):

Spenhouet:
@stn1slv Thanks for that addition. Can you resolve the merge conflict?

stn1slv:
@Spenhouet, the conflict is resolved. I merged main into the branch instead of rebasing, so your approval is still there. CI passes.

PR #293, opened 2026-08-08
fix(tables): repeat merged cell content across spanned cells

### LINKED ISSUE #292: Merged table cells (rowspan/colspan) lose their content in the exported Markdown
### Description

When a Confluence table contains merged cells (`rowspan` / `colspan`), the merged value is lost in the exported Markdown. Only the top-left position of the merged region keeps the content; every other position it covers is exported as an empty cell.

Markdown (GFM) pipe tables have no merged cells, so the exporter expands each table into a rectangular grid. The grid geometry is correct (column count and column order are right), but the content of the merged region is not carried over, so rows silently lose data that is clearly visible in Confluence.

### Steps to reproduce

Export a page containing a table where a cell is merged across two rows, e.g. the following markup (this is what the rendered `body.view` HTML looks like):

```html
<table>
  <tr><th>Env</th><th>Branch</th><th>Cluster</th></tr>
  <tr><td>Staging</td><td>release</td><td rowspan="2">Cluster A</td></tr>
  <tr><td>Test</td><td>main</td></tr>
</table>
```

In Confluence, both the *Staging* and the *Test* row visibly belong to "Cluster A".

### Actual behaviour

```markdown
| Env | Branch | Cluster |
| --- | --- | --- |
| Staging | release | Cluster A |
| Test | main |  |
```

The *Test* row lost "Cluster A". A reader of the exported vault, or any query over it (grep, Dataview, a spreadsheet import), cannot tell which cluster applies to *Test*.

The same happens for horizontal merges: a `<td colspan="3">Full width</td>` exports as `| Full width |  |  |`.

### Expected behaviour

Since Markdown cannot represent a merge, the most faithful and most useful representation is to repeat the merged value in every grid position it covers:

```markdown
| Env | Branch | Cluster |
| --- | --- | --- |
| Staging | release | Cluster A |
| Test | main | Cluster A |
```

This keeps every row self-contained and independently readable, which matters for the row-oriented tools people use on exported Markdown.

### Cause

`pad()` in `confluence_markdown_exporter/utils/table_converter.py` fills every position covered by a span with a freshly created, empty `<td>` (`make_empty_cell()`):

```python
cur.extend(make_empty_cell() for _ in range(1, cs))   # colspan positions
...
occ[(r + i, c + j)] = make_empty_cell()                # rowspan positions
```

So the grid has the right shape, but the merge origin is the only position holding content. There is currently no test coverage for `rowspan` or `colspan` at all, which is why this went unnoticed.

### Environment

- confluence-markdown-exporter 5.3.0
- Confluence Cloud

I have a fix ready and will open a PR referencing this issue.

stn1slv:
Closes #292

## What

Merged Confluence table cells (`rowspan` / `colspan`) now repeat their value in every grid position the merge covers, instead of exporting blank cells.

Before:

```markdown
| Env | Branch | Cluster |
| --- | --- | --- |
| Staging | release | Cluster A |
| Test | main |  |
```

After:

```markdown
| Env | Branch | Cluster |
| --- | --- | --- |
| Staging | release | Cluster A |
| Test | main | Cluster A |
```

## Why

Markdown pipe tables have no merged cells, so `pad()` expands each table into a rectangular grid. It filled every position covered by a span with a freshly created empty `<td>` (`make_empty_cell()`), so the geometry was right but the content survived only in the merge origin. Every other covered row or column exported as blank, silently dropping data that is plainly visible in Confluence.

Repeating the value is the most faithful representation available in GFM and keeps each row self-contained, which matters for the row-oriented tools people use on exported Markdown (grep, Dataview queries, spreadsheet imports).

### Trade-off on horizontal merges

Note that `colspan` never actually lost data before this change: the old code emitted the value once and padded the rest of the row with blanks. It is included here so that one rule covers both merge directions, but it does change output, and there is a cost worth calling out explicitly.

A full-width section row is now printed once per column, and because the repeated text widens the row, a `mixed` mode table can cross `_MAX_TABLE_LINE_LEN` and fall back to compact formatting. Measured on a 4-column table with a full-width section row: a short title keeps the table aligned at 58 characters, while `Non-production environments` reaches 121 characters and flips the whole table to compact.

```markdown
| **Env** | **Branch** | **Cluster** | **Owner** |
| --- | --- | --- | --- |
| Non-production environments | Non-production environments | Non-production environments | Non-production environments |
| Dev | development | Cluster A | Team 1 |
```

Vertical merges never have this effect, since they add no width. If you would rather keep `colspan` as it was, or repeat it only when the cell does not span the entire row, that is a small change to `pad()` and I am happy to adjust.

## How

- `pad()` fills spanned positions with the merge **origin `Tag` object itself** rather than a new empty cell.
- `convert_table()` memoizes `process_tag` per cell object, so a merged cell is converted exactly once no matter how many positions it covers. This is required, not an optimisation: cell conversion is stateful, and `_extract_uml_from_storage` advances `self._plantuml_index` on every converted PlantUML macro, so converting a repeated cell again would misalign every later diagram on the page.
- `rowspan` / `colspan` are clamped to a minimum of `1`. This is what stops `colspan="0"` (valid in HTML4, dropped in HTML5) or a negative value from repeating the cell zero times and erasing it from the row. `rowspan="0"` validly means "to the end of the row group"; treating it as a single row is a pragmatic simplification rather than a spec-accurate reading.
- The occupancy loop starts at `i = 1`; the `i == 0` entries it used to write were same-row positions the column cursor had already passed, so they were never read.
- `make_empty_cell()` is removed, since `pad()` was its only caller.

Tables without merged cells are byte-for-byte unaffected: every cell is a distinct object, so the cache always misses and the conversion sequence is identical to before.

## Tests

The repo had no `rowspan` / `colspan` coverage at all, which is why this went unnoticed. A new `TestMergedCells` class covers:

- `rowspan=2` and `rowspan=3` repeating down the covered rows
- `colspan=3` repeating across the covered columns
- a combined `rowspan=2 colspan=2` block
- `rowspan` on a `<th>` in the header row
- geometry regression: cells following a first-column span keep their column
- unparsable `colspan="abc"` and `rowspan="0"` do not drop cells
- `colspan="0"` and a negative `colspan` do not erase the cell, which is the case that actually pins the clamp
- a merged cell is passed to `process_tag` exactly once

Assertions compare whole rows in `compact` mode, so column order and column count are verified, not just substring presence.

## Verification

- `uv run pytest`: 451 passed. 7 of the 9 new tests fail on `main`, confirming the regression they pin down, and the zero/negative `colspan` test fails if the clamp is removed
- `uv run ruff check .`: clean; `ruff format --check` clean for both changed files
- End-to-end export of a real Confluence Cloud page containing a 5-column table with two cells merged across two rows: the previously blank cells now carry the merged values, and the unmerged rows are unchanged
- Horizontal merges checked separately against the markup Confluence Cloud actually emits (`<colgroup>`, `<tbody>`, `confluenceTd` classes, `<p>`-wrapped cell content), in both `compact` and `mixed` mode: partial `colspan`, `colspan` in the header row, two `colspan` cells in one row, `colspan` combined with `rowspan`, and a `colspan` wider than the header row (truncated to the header's column count, which loses nothing because every copy is identical)

### DIFF
--- a/confluence_markdown_exporter/utils/table_converter.py
+++ b/confluence_markdown_exporter/utils/table_converter.py
@@ -67,9 +67,15 @@ def _get_int_attr(cell: Tag, attr: str, default: str = "1") -> int:
 def pad(rows: list[list[Tag]]) -> list[list[Tag]]:
     """Expand rowspan and colspan into a rectangular grid of cells.
 
-    Markdown tables cannot merge cells, so every grid position a merged cell covers
-    references the merge origin itself and therefore renders the same value. The same
-    Tag object is reused on purpose, so callers can convert each origin only once.
+    Markdown tables cannot merge cells, so in a well-formed table every grid position a
+    merged cell covers references the merge origin itself and therefore renders the same
+    value. The same Tag object is reused on purpose, so callers can convert each origin
+    only once.
+
+    A ragged row, one with too few cells to reach a position reserved by an earlier
+    rowspan, still drops that reservation, because the trailing sweep only fills the
+    positions directly following the last cell of the row. Such markup is malformed and
+    renders the same way it did before merged content was carried over.
     """
     padded: list[list[Tag]] = []
     occ: dict[tuple[int, int], Tag] = {}
@@ -82,8 +88,10 @@ def pad(rows: list[list[Tag]]) -> list[list[Tag]]:
             while (r, c) in occ:
                 cur.append(occ.pop((r, c)))
                 c += 1
-            # Malformed markup can yield 0 (or a negative value); such a cell still
-            # occupies a single grid position.
+            # Every cell occupies at least one grid position: colspan="0" (valid in
+            # HTML4, dropped in HTML5) or a negative value would otherwise erase it.
+            # rowspan="0" validly means "to the end of the row group"; treating it as a
+            # single row is a pragmatic simplification, not a spec-accurate reading.
             rs = max(_get_int_attr(cell, "rowspan", "1"), 1)
             cs = max(_get_int_attr(cell, "colspan", "1"), 1)
             # Repeat the cell across the columns it spans in this row.
@@ -163,6 +171,9 @@ class TableConverter(MarkdownConverter):
         # A merged cell occupies several grid positions but must be converted only
         # once: conversion is stateful (e.g. the PlantUML macro index advances per
         # converted macro), so repeated positions reuse the first conversion.
+        # Keyed on id() and not on the Tag itself: bs4 gives Tag a structural __eq__
+        # and __hash__, so two distinct cells with identical markup would collapse
+        # into one cache entry and wrongly share a conversion.
         cell_cache: dict[int, str] = {}
         converted: list[list[str]] = []
         for row in padded_rows:
--- a/confluence_markdown_exporter/utils/table_converter.py
+++ b/confluence_markdown_exporter/utils/table_converter.py
@@ -65,7 +65,12 @@ def _get_int_attr(cell: Tag, attr: str, default: str = "1") -> int:
 
 
 def pad(rows: list[list[Tag]]) -> list[list[Tag]]:
-    """Pad table rows to handle rowspan and colspan for markdown conversion."""
+    """Expand rowspan and colspan into a rectangular grid of cells.
+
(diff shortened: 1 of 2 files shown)


stn1slv:

Spenhouet (approved):

PR #297, opened 2026-08-17
build(deps): cap atlassian-python-api below 5

### LINKED ISSUE #296: Fresh installs broken: unbounded atlassian-python-api now resolves to 5.x
### Summary

Every fresh install of `confluence-markdown-exporter` is currently broken, including pinned installs of 5.3.0, because `atlassian-python-api` is declared without an upper bound and now resolves to the new 5.x major (5.0.0 released 2026-08-15, 5.0.2 on 2026-08-16). The repository's `uv.lock` still pins 4.0.7, so CI here stays green while users installing from PyPI get 5.x.

### Symptom

A scheduled export that had been running unchanged for weeks started failing on 2026-08-17. cme reports it as an authentication problem and then opens the interactive credentials menu, which dies with `EOFError` because there is no TTY:

```
ERROR    Failed to retrieve spaces: 404 Client Error: Not Found for
         url: https://<site>.atlassian.net/wiki/wiki/api/v2/spaces?limit=1&cursor=...
ConnectionError: Confluence connection failed: 404 Client Error: Not Found
AuthNotConfiguredError: No valid authentication configured for Confluence at https://<site>.atlassian.net
...
EOFError
```

The credentials are valid. Note the doubled `/wiki/wiki/` in the URL.

Same command, same cme version, only the transitive dependency differs:

| Run | `confluence-markdown-exporter` | `atlassian-python-api` | Result |
|---|---|---|---|
| 2026-08-14 | 5.3.0 | 4.0.7 | success |
| 2026-08-17 | 5.3.0 | 5.0.2 | failure |

### Reproduction of the immediate failure

`ApiClientFactory.create_confluence` validates the connection with `get_all_spaces(limit=1)`. With 5.x the Cloud client serves this from the v2 endpoint with cursor pagination, and the relative `_links.next` path is concatenated onto a base URL that already ends in `/wiki`:

```python
from atlassian.confluence.cloud.cloud import ConfluenceCloud
# with the HTTP layer stubbed to return a first page whose
# _links.next is "/wiki/api/v2/spaces?limit=1&cursor=C":
#   page 1: https://<site>.atlassian.net/wiki/api/v2/spaces?limit=1        -> OK
#   page 2: https://<site>.atlassian.net/wiki/wiki/api/v2/spaces?...       -> 404
```

The offending join is in `atlassian/confluence_base.py`, V2 branch of `_get_paged`. I will report that separately to `atlassian-api/atlassian-python-api`.

### Why 5.x needs more than a version bump

Probing both versions with the HTTP layer stubbed, `Confluence(url=..., cloud=True)`:

| Aspect | 4.0.7 | 5.0.2 |
|---|---|---|
| `client.url` | `https://site/wiki` | `https://site` (no `/wiki` context) |
| `api_root` | `rest/api` | `wiki/api/v2` |
| `get_page_by_id` | present | `AttributeError` on the Cloud class |
| `get_attachments_from_content` | present | `AttributeError` |
| `get_page_comments` | present | `AttributeError` |
| `get_all_spaces(space_type=, space_status=, expand=)` | v1 dict, homepage expansion | v2 objects, those kwargs ignored |
| `client.get("rest/api/content/search", ...)` | `https://site/wiki/rest/api/...` | `https://site/rest/api/...` (404) |

cme depends on all of these: `Organization.from_url` uses the v1 `get_all_spaces` kwargs, `_fetch_page_ids_cql_batch` and `_fetch_comment_replies` call raw v1 paths, and attachment download builds `client.url + download_link`. The v2 API also has no CQL equivalent, so some v1 usage has to stay regardless.

### Suggested fix

Cap the dependency to the 4.x line so installs keep working, and treat the 5.x migration as separate work. PR: https://github.com/Spenhouet/confluence-markdown-exporter/pull/297

### Workaround for users right now

```
uv tool install confluence-markdown-exporter==5.3.0 --with "atlassian-python-api<5"
```

### Unrelated but adjacent

When the connection check fails, cme falls back to the interactive credentials menu even when stdin is not a TTY. In CI this converts any connection error into an `EOFError` and hides the real cause. Happy to open a separate issue for that if you think it is worth changing.

stn1slv:
Fixes #296

### What

Cap `atlassian-python-api` to `>=4,<5` in `pyproject.toml` and record the specifier in `uv.lock`. No code changes.

### Why

`atlassian-python-api` 5.0.0 was released on 2026-08-15 and is a breaking rewrite of the Confluence client. Because the dependency had no upper bound, every fresh install of cme (including pinned installs of 5.3.0) now resolves to 5.x and fails at the first Confluence call. The lockfile in this repository still pins 4.0.7, so CI does not see the breakage, but users installing from PyPI do.

What 5.x changes for cme, measured on `Confluence(url=..., cloud=True)` with the HTTP layer stubbed:

- `client.url` no longer carries the `/wiki` context, which breaks the raw v1 REST calls in `_fetch_page_ids_cql_batch` and `_fetch_comment_replies` and the attachment download that builds `client.url + download_link`.
- `get_page_by_id`, `get_attachments_from_content` and `get_page_comments` no longer exist on the Cloud class.
- `get_all_spaces` serves the v2 endpoint and ignores the v1 `space_type` / `space_status` / `expand` arguments used by `Organization.from_url`.
- Its v2 cursor pagination joins the relative `_links.next` onto a base that already ends in `/wiki`, requesting `/wiki/wiki/api/v2/spaces` and getting a 404. That is the error users hit first, in the connection check in `ApiClientFactory.create_confluence`.

The cap restores the dependency set that the lockfile already uses and that every working install has. It can be lifted once cme is migrated to the 5.x client, which is a larger piece of work: the v2 API has no CQL equivalent, so part of the v1 usage has to stay in any case.

### Verification

- `uv lock --check` passes with no re-resolution.
- `uv run pytest` passes (442 tests).
- `uv tool install confluence-markdown-exporter==5.3.0 --with "atlassian-python-api<5"` resolves `atlassian-python-api==4.0.7` and the export runs green again, which is the same constraint this PR encodes in the package metadata.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,7 +14,11 @@ authors = [
 ]
 requires-python = ">= 3.10"
 dependencies = [
-    'atlassian-python-api',
+    # atlassian-python-api 5.x is a breaking rewrite of the Confluence client:
+    # the Cloud class drops get_page_by_id, get_attachments_from_content and
+    # get_page_comments, client.url no longer carries the /wiki context, and the
+    # v2 cursor pagination builds an invalid next-page URL. Cap until we migrate.
+    'atlassian-python-api>=4,<5',
     'jmespath',
     'markdownify',
     'pydantic-settings',
--- a/uv.lock
+++ b/uv.lock
@@ -238,7 +238,7 @@ docs = [
 
 [package.metadata]
 requires-dist = [
-    { name = "atlassian-python-api" },
+    { name = "atlassian-python-api", specifier = ">=4,<5" },
     { name = "jmespath" },
     { name = "lxml", specifier = ">=6.0.2" },
     { name = "markdownify" },


stn1slv:

Spenhouet (approved):

Spenhouet:
@stn1slv thanks for that quick fix!

PR #321, opened 2026-09-03
fix(export): delete old file when a page is renamed or moved

### LINKED ISSUE #320: Renaming a page leaves the file with the old name behind
## What happens

When a page is renamed in Confluence, the next export writes it to the new path and leaves the file at the previous path in place. The export ends up with two files for one page, and the old one keeps stale content.

Renaming a parent page multiplies this, because `{ancestor_titles}` is part of the default `export.page_path`. Every descendant moves too, so the whole old subtree is duplicated.

## Steps to reproduce

1. Export a space with default settings (`skip_unchanged=true`, `cleanup_stale=true`).
2. Rename a page in Confluence from `Old Title` to `New Title`.
3. Export the same space again.

Expected: `Old Title.md` is gone, `New Title.md` is there.

Actual: both files exist. `confluence-lock.json` records only the new path, so nothing tracks the old file any more.

## Cause

`LockfileManager.remove_pages` already contains a loop that deletes the old file when a page's `export_path` changed. It never runs here, because `sync_removed_pages` returns before reaching it:

```python
unseen = LockfileManager.unseen_ids()
if not unseen:
    logger.debug("No unseen pages in lockfile — nothing to clean up.")
    return
```

A renamed page is seen during the run, since it gets exported, so it is never in `unseen_ids()`. On a full-space sync where nothing was deleted in Confluence, `unseen` is empty and the function returns early. The moved-page cleanup only fires by accident, when some unrelated page happens to be missing from that run.

The early return avoids a pointless API call, which is worth keeping. The problem is that it also skips work needing no API at all.

The page's `.comments.md` sidecar has the same problem from a different direction. It is not tracked in the lockfile and is never removed, so a rename orphans that file too when `export.comments_export` is enabled.

Version: 5.4.0

I have a fix and will open a PR.

stn1slv:
Fixes #320

## What

`sync_removed_pages` now always runs the moved-page cleanup in `LockfileManager.remove_pages`, instead of returning early whenever no page is unseen. The API existence check is still skipped when there is nothing unseen, so this costs no extra requests.

A page's `.comments.md` sidecar is now deleted along with the page, for renamed and deleted pages alike.

## Why

A renamed page is seen during the export run, so it never appears in `unseen_ids()`. On a full-space sync where nothing was deleted in Confluence, `unseen` is empty and `sync_removed_pages` returned before calling `remove_pages`. The moved-page unlink loop inside it was therefore unreachable in the common case, and the file under the old title survived. Renaming a parent page left the entire old subtree behind, because `{ancestor_titles}` is part of the default `export.page_path`.

The existing test for that loop called `remove_pages` directly, which is why it passed while the real path stayed broken.

## Guards

Making the loop run on every sync arms a delete path that previously almost never executed, so two guards ship with it.

The first covers case-only renames. `should_export` compares export paths as strings, but on a case-insensitive filesystem, which is the macOS and Windows default, re-exporting `Page` as `PAGE` writes through the same file. Unlinking the old path would then delete the content just written. `_resolves_to_same_file` skips the unlink when both paths are the same file on disk.

The second covers path reuse. If another page now occupies the old path, say after renaming `Release Notes` to `Release Notes 2025` and creating a fresh `Release Notes`, that path belongs to a live page. A path-ownership check skips those.

Unlink failures other than "file not found" are now logged rather than raised, so a file held open by another process cannot fail the command at the end of an otherwise successful export.

## Tests

Nine tests added, six of which fail on `main`:

- The regression test drives `sync_removed_pages` with no unseen pages and asserts the old file is deleted, the new file survives, and the lockfile entry is kept.
- Two tests pin the existing behaviour of skipping and running the API existence check.
- Two cover sidecar removal, on rename and on delete.
- Three guard tests: two pages swapping paths, a case-only rename, and a deleted page whose path another page has taken.

Verified end to end against a real 842-page lockfile with a renamed parent and two children. On `main` all six stale files survive. With this change all six are removed and the new files are intact.

## Known limitations, not addressed here

- Empty directories are left behind after a rename moves a subtree.
- Pages linking to the renamed page are not re-exported, since their own version and path did not change, so their relative links now point at a deleted file. Previously the stale file absorbed those links.
- A page moved between spaces keeps an entry under both space keys, because `add_page` never removes the old one.
- Attachments only move on rename if `attachment_path` is customised to include title components. The default template does not.
- `cleanup_stale=true` with `skip_unchanged=false` still does nothing, because `LockfileManager.init` returns before setting `_lock`.

## Behaviour change worth noting

Changing `export.page_path` now deletes the old tree on the next sync, since every page's path changes at once. That matches what `cleanup_stale` documents, but anyone relying on the old files lingering will notice.

### DIFF
--- a/confluence_markdown_exporter/confluence.py
+++ b/confluence_markdown_exporter/confluence.py
@@ -3438,16 +3438,18 @@ def sync_removed_pages(base_url: str) -> None:
         logger.debug("Stale page cleanup disabled — skipping.")
         return
 
+    # Renamed and moved pages are seen during the run, so they never show up as
+    # unseen. Their old files still need removing, which needs no API call.
+    deleted: set[str] = set()
     unseen = LockfileManager.unseen_ids()
-    if not unseen:
-        logger.debug("No unseen pages in lockfile — nothing to clean up.")
-        return
+    if unseen:
+        with console.status(f"[dim]Checking {len(unseen)} unseen page(s) for removal…[/dim]"):
+            deleted = fetch_deleted_page_ids(sorted(unseen), base_url)
+        if deleted:
+            logger.info("Removing %d stale page(s) from local export.", len(deleted))
+    else:
+        logger.debug("No unseen pages in lockfile — skipping existence check.")
 
-    with console.status(f"[dim]Checking {len(unseen)} unseen page(s) for removal…[/dim]"):
-        deleted = fetch_deleted_page_ids(sorted(unseen), base_url)
-
-    if deleted:
-        logger.info("Removing %d stale page(s) from local export.", len(deleted))
     LockfileManager.remove_pages(deleted)
 
 
--- a/confluence_markdown_exporter/utils/lockfile.py
+++ b/confluence_markdown_exporter/utils/lockfile.py
@@ -28,6 +28,18 @@ logger = logging.getLogger(__name__)
 LOCKFILE_VERSION = 2
 
 
+def _try_unlink(path: Path) -> bool:
+    """Delete *path*, returning True if it existed. Never raises."""
+    try:
+        path.unlink()
+    except FileNotFoundError:
+        return False
+    except OSError:
+        logger.warning("Failed to delete stale file: %s", path)
+        return False
+    return True
+
+
 class AttachmentEntry(BaseModel):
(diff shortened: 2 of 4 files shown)


stn1slv:

Issue #298, opened 2026-08-18
Page link resolution fetches full page bodies: one heavy linked page costs minutes, with no way to skip it


fenrire:
## Summary

Resolving an outgoing **page link** triggers a *full* page fetch (`Page.from_id`, 10 expands including all four body formats), even though `convert_page_link` only ever uses `title`, `id` and `export_path`. When a linked page is slow or unexportable, each link costs minutes. A single page that links to many such pages can consume an entire export budget, and there is currently no way to avoid it.

## Environment

- `confluence-markdown-exporter` 5.2.1 (observed); **code paths below verified unchanged in 5.4.0**
- `atlassian-python-api` 4.0.7 (latest 4.x; 5.4.0 pins `>=4,<5`)
- Confluence Cloud, `cme pages <urls>` batches of 50, incremental export driven by CQL

## What we observed

One page in our space (an "inventory" page, 25 KB of storage) links to **38 other pages**. Six of those are pages we already know cannot be exported (very large weekly reports, ~170 K chars of storage, which our exporter config skips via our own exclusion list).

Exporting that *one* page took **22 of 33 minutes** of a run whose other 39 spaces finished in 3 minutes combined. Timeline of a single batch:

```
17:59:54 ERROR Expecting value: line 1 column 1 (char 0)
         WARNING Could not access page id=2778955949 — treating as inaccessible
         WARNING Confluence page link (ID: 2778955949) is not accessible,
                 referenced from page 'inventory page' (ID: 2897445038)
18:01:55 (same, next link target)
18:03:56 (same, next link target)
18:05:56 ...
18:07:56 ...
```

Each failing link target costs ~120 s: Confluence takes that long to return an (empty) body for the heavy `expand` combination, `json` decoding fails, and cme then correctly falls back to "treating as inaccessible". **The end result for every one of those links is `[Page not accessible (ID: …)]` — the same string we would get for free.**

Worse, with the default connection timeout (120 s) the fetch *times out* instead of returning, and we observed a **single link target being retried for 50 minutes** until an external watchdog killed the process. That part comes from the dependency: `atlassian-python-api` mounts

```python
# atlassian/rest_client.py
retries = Retry(
    total=None,
    status=self.max_backoff_retries,
    ...
```

[...]

PR #324, opened 2026-09-04
feat(links): resolve Confluence tiny links (/wiki/x/) to pages

### LINKED ISSUE #323: Tiny links (/wiki/x/...) are exported as bare URLs instead of page links
**What happens**

A link inserted into a page by pasting the URL from "Copy link" is a tiny link, for example `https://example.atlassian.net/wiki/x/Fc1bBw`. When Confluence turns the pasted URL into an inline card, the export comes out as a bare autolink:

```
<https://example.atlassian.net/wiki/x/Fc1bBw>
```

A page link inserted through the link picker, or a pasted full URL (`/wiki/spaces/KEY/pages/123456789/Title`), is exported as `[Page Title](Relative/Path.md)`. The tiny link form keeps the absolute URL and also loses the page title as link text, because the inline card carries the URL as its text.

**Why**

In the exported HTML such an anchor looks like this:

```html
<a href="https://example.atlassian.net/wiki/x/Fc1bBw" data-card-appearance="inline" rel="nofollow">https://example.atlassian.net/wiki/x/Fc1bBw</a>
```

There is no `data-linked-resource-id` and no `pageId=` query parameter, so `convert_a` falls back to `parse_confluence_path`, which only knows the Cloud `/spaces/{key}/pages/{id}` and Server `/display/{key}/{title}` shapes. `/wiki/x/{id}` matches neither, and the anchor goes through markdownify's default conversion.

The same helper is used by `Page.from_url`, so `cme page https://example.atlassian.net/wiki/x/Fc1bBw` fails with "Could not parse page URL" as well.

**Reproduce**

1. On a Confluence Cloud page, click "Copy link", paste the URL into another page and keep it as an inline card.
2. Export the second page with `cme page <url>`.
3. The link appears as `<https://.../wiki/x/...>` in the Markdown.

**Proposed fix**

The tiny identifier is the page id itself: a little-endian 64-bit integer, base64 encoded with `/` and `+` replaced by `-` and `_`, with trailing `A` and `=` stripped. It can be decoded without an API call, so `parse_confluence_path` can recognise `[/wiki]/x/{id}` and return a reference with the page id. From there the existing `convert_page_link` path produces the relative link and the page title. I have a change ready and will open a PR.

stn1slv:
Closes #323.

**What**

`parse_confluence_path` now recognises tiny links (`[/wiki]/x/{id}`, also behind the `/ex/confluence/{cloudId}` gateway prefix) and returns a `ConfluenceRef` carrying the page id. A new `decode_tiny_link_page_id()` decodes the identifier offline: it is the page id as a little-endian 64-bit integer, base64 encoded with `/` and `+` replaced by `-` and `_`, with trailing `A` and `=` stripped. `ConfluenceRef.space_key` becomes optional because a tiny link has none. Both existing callers already check `if match.space_key`, and `convert_a` reads only `page_id`, so page links inside exported pages and `cme page <tiny url>` pick this up without further changes.

**Why**

See #323. A pasted "Copy link" URL kept as an inline card is exported as a bare `<https://.../wiki/x/...>` autolink instead of `[Title](path.md)`.

**Notes**

- Tiny links to pages in other spaces or to inaccessible pages behave the same way as full page URLs do today: `Page.from_id` is called for them and logs "Page not accessible" when it cannot read the page. Each resolved link costs one `Page.from_id` call, the same as for a full URL (see #298).
- The tiny shape is tried before the Cloud and Server shapes, and a well-formed identifier that decodes to a non-positive number returns `None` rather than falling through to the Server matcher. As a result `/x/{something}` is treated as a tiny link and no longer parses as a Server path with space key `x`. Confluence Server also serves tiny links at `/x/{id}` on the context root, which is why the `/wiki` prefix stays optional. I can change this to require `/wiki` if you prefer.
- Editor URLs (`/pages/edit-v2/{id}#anchor`) are a separate case and are not handled here.

**Tests**

Unit tests cover the decoder (including the `-`/`_` substitution, a 12-digit id, the longest 11-character identifier, and a negative result), the parser shapes, and `convert_a` with an inline-card anchor. `uv run pytest` (521 passed) and `uv run ruff check` are clean.

Docs: one bullet in `docs/features.md`.

### DIFF
--- a/confluence_markdown_exporter/api_clients.py
+++ b/confluence_markdown_exporter/api_clients.py
@@ -1,5 +1,8 @@
+import base64
+import binascii
 import logging
 import re
+import struct
 import urllib.parse
 from threading import Lock
 from threading import local
@@ -97,7 +100,8 @@ def _decode_url_part(v: str | None) -> None | str:
 
 
 class ConfluenceRef(BaseModel):
-    space_key: Annotated[str, AfterValidator(_decode_url_part)]
+    # A tiny link carries only a page id, so the space key is optional.
+    space_key: Annotated[str | None, AfterValidator(_decode_url_part)] = None
     page_id: int | None = None
     page_title: Annotated[str | None, AfterValidator(_decode_url_part)] = None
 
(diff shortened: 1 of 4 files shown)


stn1slv:

Issue #300, opened 2026-08-21
Images embedded via raw <img src> (no macro attributes) keep an absolute /download/attachments/... URL instead of a local attachment path


sebastianbraun25:
### Description
Some pages contain `<img>` tags referencing an attachment via a plain, absolute Confluence download URL (e.g. copy-pasted HTML, older editor content) instead of a recognized macro attribute. These are exported with the raw remote URL unchanged - the export isn't fully offline/portable, and `attachments_export=referenced` doesn't download the attachment either since it's never recognized as referenced.

### Root cause
`Page.Converter.convert_img()` only resolves attachments via `data-media-id`, `data-linked-resource-file-id`, `data-linked-resource-id`, `data-encoded-xml` (`ri:filename`). If none match, it falls back to the raw `href`/`src` verbatim - no fallback parses the well-known `/download/attachments/<pageId>/<filename>` URL pattern.

### Suggested fix
Add a regex fallback that extracts `pageId`/`filename` from a raw download URL, looks up the attachment via `get_attachments_from_content`, and resolves it like the other branches - ideally also making it count as "referenced".

### Environment
Confluence Server/Data Center, confluence-markdown-exporter 5.3.0. Happy to help test a fix or contribute a PR if a maintainer confirms the preferred approach.

Spenhouet:
Thanks for the suggestion. Can you put up a PR for it?

hterik:
Made an an alternative fix in https://github.com/Spenhouet/confluence-markdown-exporter/pull/325 before seeing #307. Looks like they are using similar approaches. Please take a look if it's still relevant.

PR #325, opened 2026-09-07
fix: Attach non *.drawio-images missing data-media-* attributes

hterik:
Confluence sometimes return `img` tags without `data-media-*` attributes, this was previously handled by a special case only for images with `*.drawio.png` in the name. Other attachments, and diagrams without the `drawio.png` file name were lost and got linked as absolute hrefs back to the live Confluence instance.

With this fix, the `img url` and `data-image-src` is used as a fallback to resolve the attachment, ensuring all attachments are exported to the markdown tree.

### DIFF
--- a/confluence_markdown_exporter/confluence.py
+++ b/confluence_markdown_exporter/confluence.py
@@ -2487,6 +2487,17 @@ class Page(Document):
                     if len(drawio_images) > 0:
                         attachment = drawio_images[0]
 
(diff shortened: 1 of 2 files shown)


hterik:
Possibly dupe with https://github.com/Spenhouet/confluence-markdown-exporter/pull/325, fixing https://github.com/Spenhouet/confluence-markdown-exporter/issues/300

